PR #1, opened 2019-05-30
[BB-1242] Rename project, change Postgres and Redis version

Agrendalath:
This addresses review comments.


README.rst
```diff
@@ -24,14 +24,30 @@ Moved to settings_.
 Basic Commands
 --------------
 
+Running locally with Docker
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Open a terminal at the project root and run the following for local development::
+
+    $ docker-compose -f local.yml up
```
lgp171188: @Agrendalath, passing the `-f local.yml` argument can quickly become cumbersome. So it might be a good idea to mention here that the `COMPOSE_FILE` environment variable can be set to get rid of it in every command. This is mentioned in the documentation linked below, but imho this is handy enough to mention it here itself.

Also, it is the standard Django development server port, it is probably a good idea to mention that the web application is accessible at http://localhost:8000

.circleci/config.yml
```diff
@@ -33,32 +33,41 @@ jobs:
...
             curl -L https://github.com/docker/compose/releases/download/1.19.0/docker-compose-`uname -s`-`uname -m` > ~/docker-compose
             chmod +x ~/docker-compose
             sudo mv ~/docker-compose /usr/local/bin/docker-compose
       - run:
           name: Build and push Docker image
-          filters:
-            branches:
-              only: master
           command: |
             echo $DOCKER_PASSWORD | docker login -u $DOCKER_LOGIN --password-stdin
```
lgp171188: @Agrendalath, do not forget to store these environment variables, particularly the ones which are secrets, in Vault, in case you havn't done it already.

.circleci/config.yml
```diff
@@ -33,32 +33,41 @@ jobs:
...
             docker-compose -f production.yml build
             docker-compose -f production.yml push
```
lgp171188: @Agrendalath, I don't know a lot about docker - does this push the built image(s) to a docker registry? If so, which registry is it - Docker Hub? Do the images contain any secrets baked in? If so, we cannot push the image to a public registry.

.coveragerc
```diff
@@ -1,5 +1,5 @@
 [run]
-include = sprint_management_application/*
```
lgp171188: @Agrendalath, sorry about requesting a name change. It is a huge change and a lot of work to test and verify that everything works okay.

README.rst
```diff
@@ -78,7 +94,7 @@ To run a celery worker:
 
 .. code-block:: bash
 
-    cd sprint_management_application
+    cd sprints
```
lgp171188: @Agrendalath, don't we have to run this command the one in the next line inside the container using `docker-compose`?

README.rst
```diff
@@ -78,7 +94,7 @@ To run a celery worker:
...
     celery -A config.celery_app worker -l info
 
 Please note: For Celery's import magic to work, it is important *where* the celery commands are run. If you are in the same folder with *manage.py*, you should be right.
```
lgp171188: @Agrendalath, in the following section `Sentry` is mentioned. Are we planning to use it instead of our hosted infrastructure that we use to monitor logs? If not, can we remove it from the documentation and also from the places it is used?

.circleci/config.yml
Agrendalath: Ah, good point, I'll put them there :+1:

.circleci/config.yml
Agrendalath: The images don't contain any secrets. They are loaded from `.env` files.

.coveragerc
Agrendalath: No worries, `find -type f -print0 |xargs -0 sed -i 's/sprint_management_application/sprints/g'` + a little manual work around IDE configs didn't take much time.

README.rst
Agrendalath: That's correct. I forgot to add it.

README.rst
Agrendalath: I find Sentry really handy for catching the full context of an error. It's currently set up for the server (I'll put credentials in the Vault), but I can remove it if we won't want to use it.

README.rst
```diff
@@ -42,23 +68,23 @@ Running type checks with mypy:
...
 Test coverage
 ^^^^^^^^^^^^^
 
 To run the tests, check your test coverage, and generate an HTML coverage report::
 
-    $ coverage run -m pytest
-    $ coverage html
+    $ docker-compose -f local.yml run django coverage run -m pytest
+    $ docker-compose -f local.yml run django coverage html
     $ open htmlcov/index.html
```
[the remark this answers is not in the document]
Agrendalath: @lgp171188, it's also available in, e.g., Haiku OS or Fish shell, but that's right - it's not Linux-specific. I updated the README.

PR #2, opened 2019-06-03
[BB-1243] Set up Google auth

Agrendalath:
This disables standard (credential) login/registration and enables Google auth.

## Testing instructions:
1. Run local environment with this branch.
1. Go to http://localhost:8000/admin/socialaccount/socialapp/add/ and set the social application according to the comment in the ticket.
1. Click the `Sign In` button on the main page.

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -3,5 +3,4 @@
   <component name="JavaScriptSettings">
     <option name="languageLevel" value="ES6" />
   </component>
-  <component name="ProjectRootManager" version="2" project-jdk-name="Remote Python 3.7.3 Docker Compose (django at [/home/piotr/work/OpenCraft/projects/sprints/local.yml])" project-jdk-type="Python SDK" />
 </project>
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -269,7 +269,7 @@ CELERYD_TASK_SOFT_TIME_LIMIT = 60
 # django-allauth
 # ------------------------------------------------------------------------------
 # https://django-allauth.readthedocs.io/en/latest/configuration.html
-ACCOUNT_AUTHENTICATION_METHOD = "username"
+ACCOUNT_AUTHENTICATION_METHOD = "email"
 ACCOUNT_USERNAME_REQUIRED = False
 ACCOUNT_EMAIL_REQUIRED = True
 ACCOUNT_EMAIL_VERIFICATION = "mandatory"
@@ -277,9 +277,10 @@ ACCOUNT_ADAPTER = "sprints.users.adapters.AccountAdapter"
 SOCIALACCOUNT_ADAPTER = "sprints.users.adapters.SocialAccountAdapter"
 SOCIALACCOUNT_EMAIL_VERIFICATION = False
 # Custom options for disabling login/registration.
-ACCOUNT_ALLOW_LOGIN = env.bool("DJANGO_ACCOUNT_ALLOW_LOGIN", False)
-ACCOUNT_ALLOW_REGISTRATION = env.bool("DJANGO_ACCOUNT_ALLOW_REGISTRATION", False)
+ACCOUNT_ALLOW_LOGIN = env.bool("DJANGO_ACCOUNT_ALLOW_LOGIN", True)
+ACCOUNT_ALLOW_REGISTRATION = env.bool("DJANGO_ACCOUNT_ALLOW_REGISTRATION", True)
 SOCIALACCOUNT_ALLOW_REGISTRATION = env.bool("DJANGO_SOCIALACCOUNT_ALLOW_REGISTRATION", True)
+ACCOUNT_ALLOWED_EMAIL_DOMAINS = env.list("DJANGO_ACCOUNT_ALLOWED_EMAIL_DOMAINS", default=["opencraft.com"])
 
 
 # Your stuff...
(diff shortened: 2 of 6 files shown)


lgp171188:
@Agrendalath, one thing I noticed while testing this app is that a user cannot sign out as long as he/she is signed into the Google account used to log in to the site. I don't know if this is intentional and expected, but as some one who often wears a tin-foil hat, I wouldn't want to rely on Google as the only way to sign in to the site.

sprints/users/adapters.py
```diff
@@ -13,4 +13,4 @@ def is_open_for_signup(self, request: HttpRequest):
 
 class SocialAccountAdapter(DefaultSocialAccountAdapter):
```
lgp171188: @Agrendalath, do we want any Google account to be able to login and not just the ones under the `opencraft.com` domain? If so, some code changes (and corresponding tests) will be required.

.idea/misc.xml
```diff
@@ -3,5 +3,4 @@
   <component name="JavaScriptSettings">
     <option name="languageLevel" value="ES6" />
   </component>
-  <component name="ProjectRootManager" version="2" project-jdk-name="Remote Python 3.7.3 Docker Compose (django at [/home/piotr/work/OpenCraft/projects/sprints/local.yml])" project-jdk-type="Python SDK" />
```
lgp171188: @Agrendalath, are these (this and a few other changes below) related to the Google social auth setup at all? If so, can you separate them into a different commit that makes it obvious what they are for?

config/settings/base.py
```diff
@@ -264,19 +265,21 @@
...
 ACCOUNT_EMAIL_REQUIRED = True
-# https://django-allauth.readthedocs.io/en/latest/configuration.html
 ACCOUNT_EMAIL_VERIFICATION = "mandatory"
-# https://django-allauth.readthedocs.io/en/latest/configuration.html
 ACCOUNT_ADAPTER = "sprints.users.adapters.AccountAdapter"
-# https://django-allauth.readthedocs.io/en/latest/configuration.html
 SOCIALACCOUNT_ADAPTER = "sprints.users.adapters.SocialAccountAdapter"
+SOCIALACCOUNT_EMAIL_VERIFICATION = False
+# Custom options for disabling login/registration.
+ACCOUNT_ALLOW_LOGIN = env.bool("DJANGO_ACCOUNT_ALLOW_LOGIN", False)
```
lgp171188: @Agrendalath, I would love to have this enabled.

sprints/templates/base.html
```diff
@@ -59,13 +59,10 @@
...
             {% else %}
+              {% load socialaccount %}
               <li class="nav-item">
                 {# URL provided by django-allauth/account/urls.py #}
-                <a id="sign-up-link" class="nav-link" href="{% url 'account_signup' %}">{% trans "Sign Up" %}</a>
-              </li>
-              <li class="nav-item">
-                {# URL provided by django-allauth/account/urls.py #}
-                <a id="log-in-link" class="nav-link" href="{% url 'account_login' %}">{% trans "Sign In" %}</a>
+                <a id="log-in-link" class="nav-link" href="{% provider_login_url 'google' %}">{% trans "Sign In" %}</a>
```
lgp171188: @Agrendalath, is there a way to create a login page that prompts for username/email and password and allows optionally using Google for authentication - kinda like how LMS login page is?

sprints/users/adapters.py
lgp171188: @Agrendalath, I tested this and found that it works correctly and throws an error. :+1:

Agrendalath: @lgp171188, this has been set up on the Google application side.

config/settings/base.py
Agrendalath: @lgp171188, this has been pushed accidentally with #1 after renaming the whole project. Should I remove it from here and make a separate PR for this change?

.idea/misc.xml
Agrendalath: @lgp171188, this has been pushed accidentally with #1 after renaming the whole project. Should I remove it from here and make a separate PR for this change?

sprints/templates/base.html
lgp171188: @Agrendalath, I tested this and it works! Can we disable registration since this will be exposed to the internet? I am not sure if it is possible - but can we have the registration done solely through Google and the actual login be done via either? This would require the ability to set or change the password for a user created via Google. This is just a suggestion and can be worked on in a separate ticket.

Also, a minor nit (can be fixed during further work on this project) that I noticed was that on the `/users/~update/` page, the `Update` button doesn't have any styling.

Agrendalath: @lgp171188, sure, we can just set up `DJANGO_ACCOUNT_ALLOW_REGISTRATION` to `False` on prod. We also have `ACCOUNT_ALLOWED_EMAIL_DOMAINS` now, so I don't think that open registration makes a huge difference here.

> This would require the ability to set or change the password for a user created via Google.
Makes sense. Currently it's possible to use "Forgot password" button for this.

The `/users/~update/` page will be removed in BB-1244 as we won't be using it.

Agrendalath:
@lgp171188, +1 per chat conversation. Please add the proper comment later.

lgp171188:
:+1: 
- [x] I tested the social login functionality and it works
- [x] I read through the code
- [x] ~~I checked for accessibility issues~~ NA
- [x] ~~Includes documentation~~ NA

PR #3, opened 2019-06-17
[BB-1244] Sprint planning dashboard

Agrendalath:
This creates the sprint planning dashboard.

## Testing instructions:
1. Set `JIRA_SERVER`, `JIRA_USERNAME` and `JIRA_PASSWORD` in your envs - `.env`.
1. Create `frontend/.env.local` and set `REACT_APP_GOOGLE_CLIENT_ID` there.
1. Navigate to `frontend` and run `npm i && npm start`. Check how the dashboard is working in a newly opened window.


config/settings/base.py
```diff
@@ -282,6 +284,53 @@
 SOCIALACCOUNT_ALLOW_REGISTRATION = env.bool("DJANGO_SOCIALACCOUNT_ALLOW_REGISTRATION", True)
 ACCOUNT_ALLOWED_EMAIL_DOMAINS = env.list("DJANGO_ACCOUNT_ALLOWED_EMAIL_DOMAINS", default=["opencraft.com"])
 
+# Jira configs
+JIRA_SERVER = env.str("DJANGO_JIRA_SERVER")
+JIRA_USERNAME = env.str("DJANGO_JIRA_USERNAME")
+JIRA_PASSWORD = env.str("DJANGO_JIRA_PASSWORD")
+JIRA_SPRINT_BOARD_PREFIX = env.str("DJANGO_SPRINT_BOARD_PREFIX", "Sprint - ")
+JIRA_CUSTOM_FIELDS = env.dict("DJANGO_JIRA_CUSTOM_FIELDS", default={
```
bradenmacdonald: Can these not be looked up dynamically? https://gitlab.com/opencraft/dev/jira-scripts/blob/168fa3ccf15b9b5fb59f0fe60d8e779bfdab98a4/com/opencraft/jira/planning/UserStatsComputer.groovy#L86

config/settings/base.py
```diff
@@ -282,6 +284,53 @@
...
+    "sprint": 'customfield_10005',
+    "story_points": 'customfield_10002',
+    "reviewer_1": 'customfield_10200',
+    "reviewer_2": 'customfield_10201',
+})
+JIRA_BOT_USERNAME = env.str("DJANGO_JIRA_BOT_USERNAME", "crafty")
```
bradenmacdonald: Could you please add a bit more descriptive comments to the configuration? ([example](https://gitlab.com/opencraft/dev/jira-scripts/blob/168fa3ccf15b9b5fb59f0fe60d8e779bfdab98a4/com/opencraft/jira/planning/Config.groovy#L7-24))

For example, as someone who hasn't tried this out yet, it's unclear to me what the difference between `JIRA_USERNAME` and `JIRA_BOT_USERNAME` is, or where to get the data in `JIRA_CUSTOM_FIELDS`

config/settings/base.py
```diff
@@ -282,6 +284,53 @@
 SOCIALACCOUNT_ALLOW_REGISTRATION = env.bool("DJANGO_SOCIALACCOUNT_ALLOW_REGISTRATION", True)
 ACCOUNT_ALLOWED_EMAIL_DOMAINS = env.list("DJANGO_ACCOUNT_ALLOWED_EMAIL_DOMAINS", default=["opencraft.com"])
 
+# Jira configs
+JIRA_SERVER = env.str("DJANGO_JIRA_SERVER")
```
bradenmacdonald: IMHO it's kind of confusing that the environment variable starts with `DJANGO_`. If it was all `DASHBOARD_` or just no prefix, I think it would be less confusing. `DJANGO_JIRA_SERVER` sounds like some kind of django-based proxy server that runs between this application and JIRA.

sprints/dashboard/helpers.py
```diff
@@ -0,0 +1,276 @@
...
+
+from jira import JIRA
+# noinspection PyProtectedMember
+from jira.resources import Board, GreenHopperResource, Issue, Sprint, User as JiraUser
+
+from config.settings.base import JIRA_CUSTOM_FIELDS, JIRA_PASSWORD, JIRA_REQUIRED_FIELDS, JIRA_SERVER, \
+    JIRA_SPRINT_BOARD_PREFIX, JIRA_USERNAME, SPRINT_DEFAULT_COMMITMENT, SPRINT_EPIC_DIRECTIVE, \
+    SPRINT_HOURS_RESERVED_FOR_MEETINGS, SPRINT_RECURRING_DIRECTIVE, SPRINT_REVIEW_DIRECTIVE, \
+    SPRINT_STATUS_EPIC_IN_PROGRESS, SPRINT_STATUS_EXTERNAL_REVIEW, SPRINT_STATUS_MERGED, SPRINT_STATUS_RECURRING, \
+    SPRINT_STATUS_UNFINISHED
```
bradenmacdonald: Nit: please format like

```python
from config.settings.base import (
    JIRA_CUSTOM_FIELDS,
    JIRA_PASSWORD,
    JIRA_REQUIRED_FIELDS,
    ...,
)
```
(trailing comma on every line)

This format produces much nicer diffs for future changes and is required by edX's python coding standards, which we generally use as well.

sprints/frontend/src/components/Table.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+
+const largeColumn = {width: '40%'};
+const midColumn = {width: '30%'};
+const smallColumn = {width: '10%'};
+
+
+const Table = ({list}) =>
+    <div className="table">
```
bradenmacdonald: Nit: it's usually best to use an actual HTML `<table className="table">` for tabular data; it has better accessibility properties and it allows users to copy-and-paste the data into a spreadsheet application and then use make calculations using it.

sprints/dashboard/helpers.py
```diff
@@ -0,0 +1,366 @@
...
+        """Calculate available time for the upcoming sprint."""
+        return self.goal_time - self.committed_time
+
+    def __eq__(self, other) -> bool:
+        if isinstance(other, JiraUser):
+            return self.user == other
+        return self.user == other.user
+
+    def add_invalid_issue(self, issue: DashboardIssue) -> None:
+        """Add non-estimated issue to the list of invalid issues."""
```
lgp171188: @Agrendalath, nit: is `invalid` the correct word? It is just issues with 0 time remaining.

sprints/dashboard/helpers.py
```diff
@@ -0,0 +1,366 @@
...
+        planned = self.get_bot_directive(SPRINT_RECURRING_DIRECTIVE)
+        return planned * SECONDS_IN_HOUR
+
+    def get_epic_management_time(self) -> int:
+        """Get required assignee time for managing the epic."""
+        planned = self.get_bot_directive(SPRINT_EPIC_DIRECTIVE)
+        return planned * SECONDS_IN_HOUR or SPRINT_HOURS_RESERVED_FOR_EPIC_MANAGEMENT * SECONDS_IN_HOUR
+
+
+class DashboardRow:
```
xirdneh: Not a big issue but, it seems like there's multiple concerns in this file.
There's a few helper functions and also some data access classes.
The helper functions as well as the data access classes will probably be used across the project at some point. Specially since we're talking about an abstraction layer towards Jira.
Why not moving these to a `libs/jira` (or something like that) module at the root of the project?
There could be a `libs.jira.utils` module for the functions and maybe group the classes in a `libs.jira.dashboard.models.[Dashboard, Issue, Row]` modules.

sprints/dashboard/helpers.py
```diff
@@ -0,0 +1,366 @@
...
+        """
+        Calculate goal time for this user.
+        Use default if time not specified in user's profile.
+        """
+        self.goal_time = goal - SPRINT_HOURS_RESERVED_FOR_MEETINGS * SECONDS_IN_HOUR
+
+    @property
+    def committed_time(self) -> float:
+        """Calculate summary time for the upcoming sprint."""
+        return self.current_remaining_assignee_time \
```
xirdneh: Use parenthesis to avoid `\`
```suggestion
        return (self.current_remaining_assignee_time
                + self.current_remaining_review_time
               + self.current_remaining_upstream_time
               + self.future_remaining_assignee_time
               + self.future_remaining_review_time
               + self.future_epic_management_time)
```

xirdneh:
@Agrendalath I went through the code and it looks great. I left some comments but i didn't find anything fundamentally wrong.
Great work!

sprints/dashboard/helpers.py
Agrendalath: @lgp171188, would `Unestimated` (or `Non-estimated`) be a better name for this method and for the related column on the dashboard?

.envs/.production/.django
```diff
@@ -47,3 +47,7 @@ REDIS_URL=redis://redis:6379/0
 CELERY_FLOWER_USER=kJtrEYByunMyFNPJeALMRoyxcRTInfFw
```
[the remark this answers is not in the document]
Agrendalath: These are just example credentials that are overridden on the server with `.env` file.

config/settings/base.py
```diff
@@ -71,6 +71,7 @@
     'allauth.socialaccount.providers.google',
     "rest_framework",
     "drf_yasg",
+    "corsheaders",
```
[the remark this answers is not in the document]
Agrendalath: I haven't been digging into setting this up on traefik, so I'm not sure how much changes it would require. Do you think it's worth it?

sprints/dashboard/helpers.py
Agrendalath: @xirdneh, I reworked the placement of the classes and helper methods in 16147ce35faf8ff612e3c7452ddc5b4679e60130.
Please let me know if it looks a bit cleaner than before.
I'd like to keep Jira libs out of the root of the project, because there might be some confusion with this and the `jira` library we're using. Also we won't be using Jira connection outside of the dashboard.

config/settings/base.py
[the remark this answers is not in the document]
xirdneh: @Agrendalath Not a show stopper but definitely something that needs to be done at some point for better maintainability and that way Django has one less thing to do on every request. IMHO

sprints/dashboard/helpers.py
lgp171188: @Agrendalath, `unestimated`, `non-estimated` will not be appropriate for issues with 0 remaining time.

PR #6, opened 2019-07-15
[BB-1247] Prepare production

Agrendalath:
This tweaks production settings and adds CI for automating frontend and backend deployments.

## Testing instructions:
1. Visit https://sprints.opencraft.com and check that the dashboard is working as intended.


.circleci/config.yml
```diff
@@ -54,21 +57,84 @@ jobs:
...
       - test
-      - deploy:
+
+      - build_backend:
           requires:
             - test
           filters:
             branches:
-              only: master
+              only: agrendalath/bb-1247-production
```
giovannicimolin: Don't forget to change this before merging.

config/settings/base.py
```diff
@@ -287,13 +287,6 @@
 SOCIALACCOUNT_ALLOW_REGISTRATION = env.bool("DJANGO_SOCIALACCOUNT_ALLOW_REGISTRATION", True)
 ACCOUNT_ALLOWED_EMAIL_DOMAINS = env.list("DJANGO_ACCOUNT_ALLOWED_EMAIL_DOMAINS", default=["opencraft.com"])
 
-# Social
```
giovannicimolin: Why are these settings being removed? Are we passing variables in a different way for the SSO?

config/settings/production.py
```diff
@@ -132,31 +132,40 @@ class MediaRootS3Boto3Storage(S3Boto3Storage):
...
+    "DJANGO_DEFAULT_FROM_EMAIL", default="Sprints <noreply@sprints.opencraft.com>"
 )
 # https://docs.djangoproject.com/en/dev/ref/settings/#server-email
 SERVER_EMAIL = env("DJANGO_SERVER_EMAIL", default=DEFAULT_FROM_EMAIL)
 # https://docs.djangoproject.com/en/dev/ref/settings/#email-subject-prefix
 EMAIL_SUBJECT_PREFIX = env(
-    "DJANGO_EMAIL_SUBJECT_PREFIX", default="[Sprint Management Application]"
+    "DJANGO_EMAIL_SUBJECT_PREFIX", default="[Sprints]"
 )
+# https://docs.djangoproject.com/en/dev/ref/settings/#email-backend
```
giovannicimolin: Why are we using 2 email backends? 
Are there any benefits to this? 
Will we be using Django's email backend or Anymail as default?

.circleci/config.yml
```diff
@@ -54,21 +57,84 @@ jobs:
...
+      - build_frontend:
+          requires:
+            - test
+          filters:
+            branches:
+              only: agrendalath/bb-1247-production
```
giovannicimolin: This too :grin:

.circleci/config.yml
```diff
@@ -1,4 +1,7 @@
-version: 2
+version: 2.1
+orbs:
+  aws-s3: circleci/aws-s3@1.0.11
```
giovannicimolin: First time I see orbs being used. Nice :+1:

.circleci/config.yml
```diff
@@ -54,21 +57,84 @@ jobs:
...
+            aws cloudfront create-invalidation --distribution-id $AWS_CLOUDFRONT_DISTRIBUTION_ID --paths '/*'
+
+  deploy_backend:
+    docker:
+      - image: circleci/python:3.7
+    steps:
+      - run:
+          name: Deploy backend
+          command: |
+            ssh -o StrictHostKeyChecking=no $HOST_USERNAME@$HOSTNAME "/bin/bash /sprints/deploy.sh"
```
giovannicimolin: Can you disable passing credentials to forks on the CI configuration? 
This is to ensure that our production environment will be safe, and no variables will be leaked through the CI.

Agrendalath: @giovannicimolin, it is [disabled by default](https://circleci.com/docs/2.0/oss/#security). You can check that it's disabled [here](https://circleci.com/gh/open-craft/sprints/edit#advanced-settings).

.circleci/config.yml
Agrendalath: Thanks! :)

config/settings/base.py
Agrendalath: Yep, there were some problems with the library handling social auth and we switched to a different one. I forgot to remove these settings in #3. Sorry for the noise in this PR!

config/settings/production.py
Agrendalath: Anymail (Mailgun) support was included in the generated part of this project, but, since we want to handle it differently, I made `django.core.mail.backends.smtp.EmailBackend` the default one with an option to easily switch between them. Do you think it makes sense or should we remove Anymail completely from here?

giovannicimolin: @Agrendalath I think that it doesn't hurt to leave it here! We can use it to set up SES in the future, instead of using Gmail accounts or the ops email.

PR #14, opened 2019-08-21
[BB-1603] Split Jira sprints between cells and create new sprints via dashboard

Agrendalath:
This handles cross-cell tickets on the dashboard and adds a functionality to create new sprints while ending the current ones. It currently uses a workaround for dealing with currently shared sprints. There will be a follow-up to remove temporary workarounds after ending the current sprint (please see the `TODO` comments).

## Testing instructions:
1. Set `.env` according to the ticket description.
1. Create superuser with `docker-compose -f local.yml run --rm django python manage.py createsuperuser` (the button on frontend should be visible for `staff` users only).
1. Add email to your superuser account under `http://localhost:8000/admin/account/emailaddress/`.
1. Navigate to `frontend` and run `npm i && npm start`.
1. Log into the frontend.
1. Check that the dashboard is loading properly.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -411,8 +411,6 @@ SPRINT_RECURRING_DIRECTIVE = fr"\[~{JIRA_BOT_USERNAME}\]: plan (\d+) hours per s
 SPRINT_REVIEW_DIRECTIVE = fr"\[~{JIRA_BOT_USERNAME}\]: plan (\d+) hours for reviewing this task"
 # Regex for extracting sprint number from the name of the sprint.
 # It is also used for distinguishing standard sprints from special ones (e.g. Stretch Goals).
-# TODO: Uncomment this after ending the current sprint.
-# SPRINT_NUMBER_REGEX = env.str("SPRINT_NUMBER_REGEX", r"\w+\.(\d+)")
 SPRINT_NUMBER_REGEX = env.str("SPRINT_NUMBER_REGEX", r"\w+.*?(\d+)")
 # Regex for extracting sprint staring date from the name of the sprint.
 SPRINT_DATE_REGEX = env.str("SPRINT_DATE_REGEX", r"\((.*)\)")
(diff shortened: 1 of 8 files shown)


bradenmacdonald:
@Agrendalath If we have separate sprints for each cell, doesn't that mean that cross-cell tickets (e.g. assigned to someone from Serenity with a reviewer in Bebop) won't appear on the other cell's active sprint board?

Agrendalath:
@bradenmacdonald, for cell boards we're using the following filter:
```jql
(assignee in membersOf($CELL) OR "Reviewer 1" in membersOf($CELL) OR "Reviewer 2" in membersOf($CELL) OR project = $CELL) AND (fixVersion is EMPTY OR fixVersion in unreleasedVersions()) ORDER BY cf[10010] ASC
```
For Serenity it's a bit more complex, but I skipped this part, as it's not relevant here. You can view this under `Board -> Configure -> General -> Filter Query`.

I haven't found a place where the specific sprint would be bound to the board, but I might've missed something.

bradenmacdonald:
@Agrendalath I believe that if two different active sprints match the board filter, JIRA shows a dropdown that lets people switch between viewing the sprints. But it doesn't let you view both at the same time. I could be wrong, but that's how I remember seeing it work, and I think it could be pretty inconvenient here. I guess there's no way to know until we try it out, but I would recommend trying it out for one sprint before we commit permanently to using separate sprints.

Agrendalath:
@bradenmacdonald, thank you for the warning, I didn't know about this. I've found only one [screenshot](https://www.mogital.com/wp-content/uploads/2015/11/screen1.png) presenting this, but it seems to be from 2015, so it might be handled differently now. Trying it for one sprint sounds reasonable.

bradenmacdonald:
@Agrendalath  Actually it looks like there is an "All sprints" view:

Screen Shot 2019-08-21 at 10 39 00 AM

Agrendalath:
@bradenmacdonald, awesome! Thanks for checking this.

PR #20, opened 2019-09-09
[BB-1620] Prepare API for billable/non-billable hours

Agrendalath:
This introduces the sustainability dashboard API. It returns the data in the following format:
```js
{
    "billable_accounts": [
        {
            "key": "ACCOUNT_KEY",
            "name": "ACCOUNT NAME",
            "overall": 10.0,
            "by_cell": {
                "Cell 1": 6.0,
                "Cell 2": 4.0
            },
            "by_person": {
                 "User 1": 1.0,
                 "User 2": 5.0,
                 "User 3": 4.0
            }
        },
    ]
    "non_billable_accounts": [
        …
    ],
    "non_billable_responsible_accounts": [
        …
    ]
}
```

## Testing instructions:
1. Set `.env` file.
1. Create superuser with `docker-compose -f local.yml run --rm django python manage.py createsuperuser`.
1. Log into the admin account `http://localhost:8000/admin/`.
1. Execute `http://localhost:8000/sustainability/dashboard/?from=2019-07-01&to=2019-08-01` in the browser. <s>Currently it takes about 40s, but I'm trying to make this async now.</s> After adding `multiprocessing` it takes around 14s.
1. Check if the values for your name look fine for these accounts.


sprints/sustainability/utils.py
```diff
@@ -0,0 +1,23 @@
...
+def split_accounts_into_categories(accounts: List[Account]) -> Dict[str, List[Account]]:
+    """
+    Converts `List[Account]` into the following format: `Dict[category: str, List[Account]]`.
+    """
+    result: Dict[str, List[Account]] = {}
+    for account in accounts:
+        try:
+            category = getattr(account, 'category').key
+        except AttributeError:
+            category = ''
```
viadanna: Will we handle this case in code? Ideally, all accounts should have a category, no?

Agrendalath: That's right. The only useful case could be debugging. I just thought it would be cleaner to do this with catching one exceptions instead of nesting `if`s for handling unexpected cases (no category, no key (if it's possible)).

sprints/sustainability/models.py
```diff
@@ -0,0 +1,92 @@
...
+    def generate_expenses(self, accounts: List[Account]) -> List[SustainabilityAccount]:
+        """Generates aggregated worklogs for each account with `SustainabilityAccount`."""
+        args = [
+            ({'id': account.id, 'key': account.key, 'name': account.name}, self.cell_names, self.from_, self.to)
+            for account in accounts if account.status == 'OPEN'
+        ]
+
+        # Use multiprocessing for parallel API requests for accounts
+        pool = Pool(processes=settings.MULTIPROCESSING_POOL_SIZE)
+        results = [pool.apply_async(SustainabilityAccount, args=arg) for arg in args]
```
viadanna: Nice! I love this package :)

PR #21, opened 2019-09-20
[BB-1621] Prepare frontend for billable/non-billable hours

Agrendalath:
This introduces the frontend for the sustainability dashboard.
I made a change to the original design by removing the `select` box for choosing overall/cell/person stats, because using it was very unintuitive. Instead, the sustainability dashboard is aware of the sprint board's current view (whether it's showing cells/cell's board/person's board). Therefore, when you click on the cell's name, the sustainability dashboard recalculates its data for displaying cell-related data only.

All calculations here are performed on the frontend side and stored in the Local Storage, so the navigation is smooth. To reload the sustainability dashboard you need to refresh the whole page.

## Testing instructions:
1. Set `.env` file.
1. Create frontend/.env.local and set REACT_APP_GOOGLE_CLIENT_ID there.
1. Navigate to frontend and run npm i && npm start.
1. Log in.
1. Check the Sustainability dashboard for overall/cells/yourself and see if the data look good.
1. Change `From` and `To` dates and recheck one of the cases mentioned above.


sprints/sustainability/models.py
```diff
@@ -85,8 +85,8 @@ def generate_expenses(self, accounts: List[Account]) -> List[SustainabilityAccou
         ]
 
         # Use multiprocessing for parallel API requests for accounts
-        pool = Pool(processes=settings.MULTIPROCESSING_POOL_SIZE)
-        results = [pool.apply_async(SustainabilityAccount, args=arg) for arg in args]
-        output = [p.get() for p in results]
+        with Pool(processes=settings.MULTIPROCESSING_POOL_SIZE) as pool:
+            results = [pool.apply_async(SustainabilityAccount, args=arg) for arg in args]
+            output = [p.get() for p in results]
```
Agrendalath: This is not strictly related to frontend, but without closing the processes properly we won't be able to run too many queries.

lgp171188:
> Set .env file.
@Agrendalath, it has been a while since I tested this application. Do we have the list of variables/secrets to set in this file saved somewhere?,

lgp171188:
> @Agrendalath, it has been a while since I tested this application. Do we have the list of variables/secrets to set in this file saved somewhere?,

@Agrendalath, ping!

Agrendalath:
@lgp171188, I posted the envs on the ticket.

frontend/src/components/sustainability/SustainabilityBoard.js
```diff
@@ -0,0 +1,142 @@
...
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            startDate: this.props.sustainability.startDate,
+            endDate: this.props.sustainability.endDate,
+        };
+    }
+
+    // Because everyone counts months from 0.
```
giovannicimolin: :stuck_out_tongue:

sprints/sustainability/models.py
giovannicimolin: Nice fix :+1:

frontend/src/components/sustainability/SustainabilityBoard.js
```diff
@@ -0,0 +1,142 @@
...
+            data.billable = accounts.billable_accounts.by_person[id];
+            data.non_billable = accounts.non_billable_accounts.by_person[id];
+            data.non_billable_responsible = accounts.non_billable_responsible_accounts.by_person[id];
+        } else {
+            data.billable = accounts.billable_accounts.overall;
+            data.non_billable = accounts.non_billable_accounts.overall;
+            data.non_billable_responsible = accounts.non_billable_responsible_accounts.overall;
+        }
+        data.non_billable_total = data.non_billable + data.non_billable_responsible;
+        data.responsible_ratio = data.non_billable_responsible / data.billable * 100;
```
giovannicimolin: This should be:
```suggestion
        data.responsible_ratio = data.non_billable_responsible / (data.billable + data.non_billable_responsible) * 100;
```

We want the ratio between non billable and total amount of hours to be >25%.

frontend/.env
```diff
@@ -1,3 +1,4 @@
 REACT_APP_API_BASE=http://localhost:8000
 REACT_APP_GOOGLE_CLIENT_ID=SET_THIS
 REACT_APP_JIRA_URL=https://jira.server
+REACT_APP_MAX_NON_BILLABLE_TO_BILLABLE_RATIO=.25
```
giovannicimolin: This should be 0.4 with the new Non-billade/Billable metric.
`1 h (Non-billable) / 2.5 h (Billable) = 0.4` as mentioned [on the handbook](https://handbook.opencraft.com/en/latest/cell_budgets/)

frontend/src/App.js
```diff
@@ -77,6 +78,8 @@ class RootContainerComponent extends Component {
                         <Route component={Login}/>
                     </Switch>
                 </BrowserRouter>
+                <br/>
```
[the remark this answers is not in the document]
Agrendalath: Good catch, that was a quick fix. I replaced it with `margin-top`.

frontend/.env
Agrendalath: Thank you!

PR #29, opened 2019-10-07
[BB-1717] create FF and DD tickets at the end of the sprint

Agrendalath:
This implements creating FF and DD tickets automatically at the end of the sprint.

## Testing instructions 
(this will actually create new tickets, so please remember to archive them; you can see BB-1751 - BB-1756 that were created after I ran the following test)
1. Add the following under line `sprints/dashboard/tasks.py:137`:
	```python
	cell_dict = {
	    'key': cell.key,
	    'name': cell.name,
	    'board_id': cell.board_id,
	}
	future_next_sprint = get_next_sprint(sprints, next_sprint)
	create_role_issues_task.delay(cell_dict, future_next_sprint.id, get_sprint_number(future_next_sprint))
	```
1. Set frontend/.env.local file (you can find it in a comment on the ticket).
1. _Optional (this will speed up testing, as the accounts' rework is not merged into the master yet)_: remove lines 82 and 83 from `frontent/src/App.js`.
1. Navigate to frontend and run `npm i && npm start`.
1. Log in.
1. Go to the cell's dashboard and press `Complete Sprint` button.


xitij2000:
@Agrendalath I don't see any tests. Could you add some?
Also, the system for fetching rotations seems to be quite brittle. Perhaps that should be handled by the app itself instead of being read from google spreadsheets?  The rotations spreadsheet can then be retired. The app will just need a list of cell members, an potentially the ability to add overrides for particular sprints.

Agrendalath:
@xitij2000, all this data actually rely on the external APIs and its external state, so as long as we don't have full e2e test setup (with 1:1 Jira dev environment) I'm not sure what we can reliably test for this PR (except the simple JQL generation, which is not prone to external changes). 


> Also, the system for fetching rotations seems to be quite brittle. Perhaps that should be handled by the app itself instead of being read from google spreadsheets? The rotations spreadsheet can then be retired. The app will just need a list of cell members, an potentially the ability to add overrides for particular sprints.

These spreadsheets allow us to view future assignments and plan holidays ahead by resolving conflicting duties. Reworking this into the sprints would be a huge task.

PR #38, opened 2019-11-18
[BB-1817] Sustainability follow-ups

Agrendalath:
This includes the following fixes:
1. Fix date format (https://xkcd.com/1179/).
2. Fix timezone bug.
3. Add 1 decimal point to `YTD Spent` that are under 1h.
4. Fix invalid time on user board when the same person is the assignee and reviewer of the ticket.
5. Remove specified prefixes from the account names.
6. Rework sustainability table (described in the ticket).
7. Add accounts filtering by period, add `Period Spent` column.
8. Make the tables more concise by reducing padding.
9. Add tooltips.

## Testing instructions 
1. Set `.env` file (you can find it in a comment on the ticket).
1. Start the backend with `docker-compose -f local.yml up --build`.
1. Set frontend/.env.local file (you can find it in a comment on the ticket).
1. Navigate to frontend and run `npm i && npm start`.
1. Log in.
1. Check the changes mentioned above.

PR #39, opened 2019-11-19
WIP: [SE-1702] Small fixes

clemente:
Included:
- *Allow overriding the default postgres variables from .env*. Otherwise, only the settings from `.envs/.local/.postgres` are read, and the database will only be created with the `postgres_user` user instead of the one requested by `POSTGRES_USER` in `.env`
- more to come


clemente:
This was already cherry-picked in https://github.com/open-craft/sprints/pull/38
Closing.

PR #41, opened 2019-12-17
[BB-1624] Send ratio/budget alerts and calculate cell budgets

Agrendalath:
This adds periodic alerts for the budget/sustainability overflows.
It also changes cell budgets to be counted from cell-specific keys instead of cell members, so if somebody from SE logs time on BB ticket, this time is accounted to BB.

## Testing instructions 
1. (**Running this will take around 962 seconds**, so you can just leave it running in the background) Go to Django shell: `docker-compose -f local.yml exec django python manage.py shell_plus` and run:
   ```
      from sprints.sustainability.tasks import *
      validate_worklog_cache.delay(long_term=True, force_regenerate=True)
   ```
1. Navigate to frontend and run `npm i && npm start`.
1. Log in.
1. To test the cell budgets:
a) change both `From` and `To` dates to `2019/11/26`,
b) check `Period Spent` of Automation account for [Bebop](http://localhost:3000/board/26) (it should be 3),
c) check that the same as above for [Daniel](http://localhost:3000/board/24/user/Daniel%20Clemente%20Laboreo) is also 3.
1. To test notifications:
a) log into Django admin,
b) add [cell](http://localhost:8000/admin/sustainability/cell/) for `Bebop`,
c) add [budget](http://localhost:8000/admin/sustainability/budget/) for any account that has non-zero `YTD Spent`,
d) go to Django shell: `docker-compose -f local.yml exec django python manage.py shell_plus` and run:
      ```
         from sprints.sustainability.tasks import *
         send_email_alerts()
      ```
   e) check that the emails are outputted.


frontend/.env
```diff
@@ -4,4 +4,8 @@ REACT_APP_JIRA_URL=https://jira.server
 REACT_APP_MAX_NON_BILLABLE_TO_BILLABLE_RATIO=.4
 REACT_APP_MAX_NON_BILLABLE_TO_BILLABLE_CELL_RATIO=.3  # https://forum.opencraft.com/t/sustainability-dashboard/431/15
 REACT_APP_COMPANY_NAME=OpenCraft
-REACT_APP_ACCOUNT_STRIP_NAMES=[]  # If multiple names are provided, stripping is performed for each of them, honoring their order.
+# If multiple names are provided, stripping is performed for each of them, honoring their order.
```
Agrendalath: Local tests were breaking because of this.

frontend/package.json
```diff
@@ -12,7 +12,7 @@
     "react-google-login": "^5.0.5",
     "react-redux": "^7.1.0",
     "react-router-dom": "^5.0.1",
-    "react-scripts": "3.1.1",
+    "react-scripts": "^3.3.0",
```
Agrendalath: Patched some vulnerabilities.

sprints/sustainability/utils.py
```diff
@@ -43,6 +53,78 @@ def generate_month_range(start: str, end: str) -> Generator[Tuple[datetime, date
...
+                issues.update(new_issues)
+                cache.set(issues_key, issues, timeout)
+
+            worklogs = cache.get(worklogs_key) or {}
+            worklogs.update(new_worklogs)
+            cache.set(worklogs_key, worklogs, timeout)
+
+    return worklogs
+
+
```
Agrendalath: I don't like this approach, but that was the only thing I could think of, due to limited API.

sprints/dashboard/tasks.py
```diff
@@ -237,17 +238,18 @@ def complete_sprint_task(board_id: int) -> None:
...
             # Ensure that the next sprint exists. If it doesn't exist, create it.
-            future_next_sprint = get_next_sprint(sprints, next_sprint)
-            if not future_next_sprint:
-                create_next_sprint_task(board_id)
-                future_next_sprint = get_next_sprint(sprints, next_sprint)
+            # Get next sprint number for creating role tasks there.
+            if future_next_sprint := get_next_sprint(sprints, next_sprint):
+                future_next_sprint_number = get_sprint_number(future_next_sprint)
+            else:
+                future_next_sprint_number = create_next_sprint_task(board_id)
```
Agrendalath: That fixes a bug that we encountered once. After some optimizations we started using cached sprints to speed things up, but forgot to refresh it here.

requirements/production.txt
```diff
@@ -5,7 +5,7 @@
 gunicorn==19.9.0  # https://github.com/benoitc/gunicorn
 psycopg2==2.8 --no-binary psycopg2  # https://github.com/psycopg/psycopg2
 Collectfast==0.6.2  # https://github.com/antonagestam/collectfast
-sentry-sdk==0.8.0  # https://github.com/getsentry/sentry-python
+sentry-sdk==0.13.5  # https://github.com/getsentry/sentry-python
```
Agrendalath: It's good to have Redis integration for Sentry :)

frontend/.env
```diff
@@ -4,4 +4,8 @@ REACT_APP_JIRA_URL=https://jira.server
...
+REACT_APP_ACCOUNT_STRIP_NAMES=[]
```
Agrendalath: Local tests were breaking, because the comment was being caught as a part of the `REACT_APP_ACCOUNT_STRIP_NAMES` env variable.

sprints/sustainability/models.py
```diff
@@ -95,7 +108,8 @@ def calculate_budgets(self, start_date: datetime, end_date: datetime) -> None:
         # Calculate available budget for the next sprint.
         daily_budget: Dict[int, float] = {
             now.month: self.calculate_workday_budget(year, now.month, self.budgets[now.month - 1]),
-            now.month + 1: self.calculate_workday_budget(year, (now.month + 1) % 12, self.budgets[now.month % 12]),
+            (now.month + 1) % 12:
```
Agrendalath: This is weird, because I cannot reproduce it locally, but it's broken on prod. I tried debugging the prod and this should work.

sprints/dashboard/libs/jira.py
```diff
@@ -107,6 +112,25 @@ def report(self, from_: str, to: str) -> Report:
...
+
+        Limited to 1000 worklogs. Source:
+        https://developer.atlassian.com/cloud/jira/platform/rest/v3/?utm_source=%2Fcloud%2Fjira%2Fplatform%2Frest%2F&utm_medium=302#api-rest-api-3-worklog-list-post
+        """
+        aggregated_worklogs: List[Dict] = []
+        for chunk in chunks(worklogs, 1000):
+            r_json = self._session.post(
+                url=self._get_url('worklog/list', self.API_V2),
+                data=json.dumps({'ids': chunk}),
+            )
```
viadanna: Can these queries be crashing Jira?

sprints/sustainability/migrations/0002_auto_20191217_1049.py
```diff
@@ -0,0 +1,31 @@
+# Generated by Django 2.2.7 on 2019-12-17 10:49
```
viadanna: Can you rename this migration to something more explicit?

sprints/sustainability/models.py
viadanna: Not sure what you meant here but could it be that the key didn't `% 12` so it could go over 12?

sprints/sustainability/models.py
```diff
@@ -239,10 +273,43 @@ def _fetch_accounts_chunk(from_: str, to: str) -> Dict[str, Dict]:
...
+            sustainability[project] = responsible_hours
+
+        return sustainability
+
+    def get_accounts_remaining_time(self) -> Dict[str, float]:
+        """Retrieve time left of all fetched accounts."""
+        accounts: Dict[str, float] = {}
+        for category in settings.TEMPO_ACCOUNT_TRANSLATE.values():
+            for account in getattr(self, category):
+                accounts[account.name] = accounts.get(account.name, 0) + account.ytd_overall - account.ytd_goal
```
viadanna: This could be simplified by using a `defaultdict(float)` so just an `accounts[account.name] += ...` is necessary. Just a suggestion though, feel free to ignore.

sprints/sustainability/tasks.py
```diff
@@ -56,3 +67,71 @@ def validate_worklog_cache(long_term=True, force_regenerate=False) -> bool:
...
+    end_date = today.replace(day=last_day_of_month)
+    start_str = start_date.strftime(settings.JIRA_API_DATE_FORMAT)
+    end_str = end_date.strftime(settings.JIRA_API_DATE_FORMAT)
+    dashboard = SustainabilityDashboard(start_str, end_str)
+
+    # Send alerts for cells.
+    max_percentage_ratio = settings.SUSTAINABILITY_MAX_NON_BILLABLE_TO_BILLABLE_CELL_RATIO * 100
+    _send_email_alert(
+        dashboard.get_projects_sustainability(),
+        lambda k, v: v >= max_percentage_ratio and Cell.objects.filter(name=k).exists(),  # Ignore non-set projects.
```
viadanna: Instead of this cryptic line, what do you think of filtering the items beforehand and just passing to `_send_email_alert` whatever needs to be emailed?

This would make this behavior more explicit, easier to understand and remove the filtering from the function that send emails.

sprints/sustainability/utils.py
viadanna: Yeah, took me a while to follow this code. Not much to do with the limited API and the need to keep the cache in sync.

One idea is checking if [webhooks](https://developer.atlassian.com/server/jira/platform/webhooks/) would be able to push all updates to sprints so it does not need to pull anything. This would make sure the cache is always in sync and greatly reduce the queries and resource usage in Jira.

sprints/dashboard/libs/jira.py
Agrendalath: It's public Jira API, so we can assume it's stable. I tested this locally for all worklogs since 2015 (which is also included in the testing instructions in point 5) and it worked fine.

sprints/sustainability/utils.py
Agrendalath: Oh, that's a nice idea! Could you please add it to the forum thread as an idea of improving things?
However I see one nit with this approach:
> To register a webhook, you can use any of the following methods:
> 
>     - Jira administration console.
>     - Jira REST API (note that the user must have the Jira Administrators global permission).

sprints/sustainability/models.py
Agrendalath: Yeah, that was the point. It shouldn't and I cannot reproduce it, but [that has actually happened](https://sentry.io/organizations/opencraft/issues/1391629621/?project=1463533&query=url%3Ahttp%3A%2F%2Fapi.sprints.opencraft.com%2Fsustainability%2Fdashboard%2F). We don't have too much time here, so I just patched this without time-consuming digging.

sprints/sustainability/models.py
Agrendalath: Good point, I did this. I couldn't do it everywhere, though, because some changes would require upfront cache invalidation for compatibility.

sprints/sustainability/tasks.py
Agrendalath: Good idea to separate this. Done.

Agrendalath:
@viadanna, thank you for the review. I addressed your comments.

Agrendalath:
@viadanna, did the step **5** return `True` for you? Because this is what seems to be breaking in the second issue. Sorry for not mentioning this explicitly. Could you please run it with Celery: `validate_worklog_cache.delay(True, True)` and check the job's status with [Flower](http://localhost:5555/tasks) (you can find `CELERY_FLOWER_*` credentials in `.env` file)?

For the first one, though, it doesn't seem to be related to this PR, as no part of the `Sprint Dashboard` has changed. I'm trying to always make data backwards-compatible, but there were many changes since your last review here, so one pair of versions may be incompatible. Could you please clear your Local Storage?

PR #45, opened 2020-03-06
[BB-2184] Calculate overhead

Agrendalath:
## What it is
This changes the way we're calculating overhead. Previously only "year-to-date" overhead was calculated. Now we also have "Period Goal" column, which represents the goal budget between chosen `From` and `To` dates (both inclusive).
The "period overhead" is now calculated into sustainability. It is added to "Non-billable cell hours" and subtracted from "Billable cell hours".

## How it works
Under the hood, the whole mechanism for calculating budgets has been rewritten to be more generic and readable. Instead of considering months between dates, we're analyzing pairs of existing budgets. The new way for calculating budgets also fixes a bug related to calculating YTD goal between multiple years.
There is also a subtle change in calculating partial budgets - previously only workdays were taken into account - so the monthly budget was divided by number of working days in a month and then multiplied by number of the working days considered in a period that is not a full month. For making code more readable, calculations easier to verify and because of [this change](https://gitlab.com/opencraft/documentation/public/-/merge_requests/61), the monthly budget is now divided into number of days in the month and multiplied by all days in a period.

## Testing instructions 
1. Download fixtures from the ticket. Upload and install them with:
      ```bash
      docker cp /tmp/budgets.json sprints_django_1:/tmp/
      docker-compose -f local.yml exec django python manage.py loaddata /tmp/budgets.json
      ```
1. Log in.
1. Check for some periods (it can be a current day for quick check) that the overhead is calculated properly and is included in sustainability in a way that was described in the first section here.

### DIFF
--- a/sprints/sustainability/models.py
+++ b/sprints/sustainability/models.py
@@ -1,9 +1,6 @@
 import calendar
+import datetime
 from collections import defaultdict
-from datetime import (
-    datetime,
-    timedelta,
-)
 from multiprocessing.pool import ThreadPool
 from typing import (
     Dict,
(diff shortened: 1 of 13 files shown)


sprints/sustainability/utils.py
```diff
@@ -40,17 +40,26 @@ def split_accounts_into_categories(accounts: List[Account]) -> Dict[str, List[Ac
...
     """Generates months between `start` and `end` dates."""
     start_date = parse(start)
     end_date = parse(end)
 
     current_date = start_date.replace(day=1)  # First day of each month
-    while current_date < end_date:
-        last_day_of_month = calendar.monthrange(current_date.year, current_date.month)[1]
-        last_date_of_month = current_date.replace(day=last_day_of_month)
+    while current_date <= end_date:
+        last_date_of_month = current_date + relativedelta(day=31)
```
giovannicimolin: Months don't always have 31 days.
Won't this wrongly take into account a few days from the following month if the month has 30 or less days?

Agrendalath: `relativedelta` does `max(31, last_day_of_the_month)` :)

It actually implements more handy things, like "Last friday in this month." [which you can grep in these examples](https://dateutil.readthedocs.io/en/stable/examples.html).

giovannicimolin: Woah, nice! That's super handy.

Agrendalath: One more subtle thing here about singulars and plurals for `relativedelta` - it might be handy in other parts of this review. Singulars **replace** the part within a range, while plurals **calculate the actual delta**:
```
year, month, day, hour, minute, second, microsecond:
    Absolute information (argument is singular); adding or subtracting a
    relativedelta with absolute information does not perform an arithmetic
    operation, but rather REPLACES the corresponding value in the
    original datetime with the value(s) in relativedelta.

years, months, weeks, days, hours, minutes, seconds, microseconds:
    Relative information, may be negative (argument is plural); adding
    or subtracting a relativedelta with relative information performs
    the corresponding arithmetic operation on the original datetime value
    with the information in the relativedelta.
```

sprints/sustainability/models.py
```diff
@@ -309,54 +366,3 @@ def get_accounts_remaining_time(self) -> Dict[str, float]:
...
-    def __str__(self):
-        return f"{self.name}: {self.date.strftime('%Y-%m')}"
-
-
-class Account(models.Model):
-    """
-    Stores email addresses for sending notifications about problems with the budgets of the account.
-    """
-    name = models.CharField(max_length=255, help_text="Account's name.")
-    alert_emails = ArrayField(
```
giovannicimolin: @Agrendalath How did we replace this? Are some emails still being notified?

Agrendalath: I just had to move `Budget` model to the top of this file, because `typing` was not detecting it from the bottom. I moved `Account` and `Cell` as well to have the consistency of class definitions.

sprints/sustainability/models.py
```diff
@@ -13,21 +10,78 @@
...
+
+class Cell(models.Model):
+    """
+    Stores email addresses for notifications about problems with the sustainability of the project.
+
+    If the project is not added, notifications will not be sent for it.
+    """
+
+    name = models.CharField(max_length=255, help_text="Cell's name.")
+    alert_emails = ArrayField(
```
giovannicimolin: @Agrendalath Just found it here again. Sorry for the noise.

PR #46, opened 2020-03-06
[BB-2186] Support special vacation

Agrendalath:
## What it is
This adds support for scheduling [special vacation](https://handbook.opencraft.com/en/latest/vacation/#special-vacation) - reduced time off.

## How it works
Due to many mistakes and oversights we relaxed regex to treat any event starting with member's name as the vacation. Additionally, if it's followed with any word and then a number after which comers letter "h", it is treated as reduced time off - **the provided number of hours indicates user's availability per day**. Some working examples:
- Piotr off
- Piotr away
- Piotr reduced 4 h only
- Piotr reduced time off (4h/day)

## Testing instructions 
1. Log in.
1. Create some test events and see if they appear on the dashboard in `Vacation` column.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -539,7 +539,7 @@ GOOGLE_API_CREDENTIALS = {
 # By default we're using f"{name} off" format, which works fine with `name` being at least user's first name.
 # CAUTION: we're not checking for duplicated names, so in case we'll have two people with the same first name,
 #          both of them will need to provide the full name in the calendar.
-GOOGLE_CALENDAR_VACATION_REGEX = env.str("GOOGLE_CALENDAR_VACATION_REGEX", r"(\w+) off")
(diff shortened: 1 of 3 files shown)


config/settings/base.py
```diff
@@ -607,4 +607,4 @@
 
 # Pool size for making parallel API requests
 MULTIPROCESSING_POOL_SIZE = env.int("MULTIPROCESSING_POOL_SIZE", 32)
-MULTIPROCESSING_TIMEOUT = env.int("MULTIPROCESSING_TIMEOUT", 32)
+MULTIPROCESSING_TIMEOUT = env.int("MULTIPROCESSING_TIMEOUT", 128)
```
Agrendalath: It's just an unrelated side change, as 32 is a low default value.

config/settings/base.py
```diff
@@ -539,7 +539,7 @@
 # By default we're using f"{name} off" format, which works fine with `name` being at least user's first name.
```
giovannicimolin: @Agrendalath Can you update this comment to reflect on the new regex behaviour?

Agrendalath: Good catch!

PR #55, opened 2020-09-04
[BB-2908] Fix calculating spreadsheet column names

Agrendalath:
## What it is
This chages the way we are calculating Excel-like column names for Google Spreadsheets, as the previous implementation was buggy for letters after `Z`. It also adds some tests to ensure this won't happen again.

## Testing instructions 
1. Open [this spreadsheet](https://docs.google.com/spreadsheets/d/1Xlpt0WGanH6v4BgmqgZJ8bgHVEBlmvyb7F-9sRxtInk/edit#gid=1636833065) on the `Bebop Commitments` tab and scroll to the last column on the right.
1. Set `.env` (you can find it in a comment on the ticket).
1. Build the backend with `docker-compose -f local.yml build`.
1. Run migrations with `docker-compose -f local.yml run django python manage.py migrate`.
1. Open Django shell with `docker-compose -f local.yml run django python manage.py shell_plus`.
1. Run the following commands there:
   ```python
   from sprints.dashboard.tasks import upload_commitments_task
   upload_commitments_task(26, 'Bebop')
   ```
1. After the command completes there should be a new column right after the last one on the spreadsheet from the point 1 of this instruction.

### DIFF
--- a/sprints/dashboard/tests/test_utils.py
+++ b/sprints/dashboard/tests/test_utils.py
@@ -6,6 +6,7 @@ from django.test import override_settings
 
 from sprints.dashboard.tests.helpers import does_not_raise
 from sprints.dashboard.utils import (
+    _column_number_to_excel,
     _get_sprint_meeting_day_division_for_member,
     extract_sprint_id_from_str,
     extract_sprint_name_from_str,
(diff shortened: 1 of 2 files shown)


aayushagra (approved):
:+1:

- [x] I tested this: ran all the steps in testing instructions and verified new columns are being created
- [ ] I read through the code
- [ ] I checked for accessibility issues
- [ ] Includes documentation
- [ ] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.

PR #58, opened 2020-11-03
[BB-3130] Allow hiding spillovers and flagged tickets

Agrendalath:
## What it is
This contains the following changes:
1. Adds a toggle for ongoing work (spillovers). This is useful to for planning the next sprint upfront by making an assumption that all work from the current sprint will be completed on time.
1. Adds a toggle for flagged tickets. If tickets are preassigned and flagged, then the assignees can use this option to determine whether they need to pass on some tickets.
1. Nit: fixes `key` attributes of unestimated tickets in `Table.js`.

image
image

**Note:** while testing, you can see that the results might sometimes be off by 1 due to rounding.

## Testing instructions 
1. Set `.env` and `frontend/env.local` (you can find them in a comment on the ticket).
1. Comment out [these two lines](https://github.com/open-craft/sprints/blob/agrendalath/bb-3130-hide_flagged_tickets/frontend/src/App.js#L87-L88). They're irrelevant to the scope, but you won't be able to render these dashboards without additional Jira permissions and I didn't test whether it won't break the whole page, as I have many things cached.
1. Register new account and get the verification link with `docker-compose -f local.yml logs django`.
1. Log in and navigate to cell's board.
1. Use the spillover checkbox and check that the numbers make sense with this enabled and disabled.
1. Flag some ticket from the next sprint and check that the toggle for flagged tickets produces correct results.
1. Go to the user's dashboard and see that the flagged ticket has yellow background.

### DIFF
--- a/frontend/src/components/sprint/UserTable.js
+++ b/frontend/src/components/sprint/UserTable.js
@@ -31,7 +31,7 @@ const UserTable = ({list, username}) =>
         </thead>
         <tbody>
         {list.map(item =>
-            <tr key={item.key} className={"table-row"}>
+            <tr key={item.key} className={"table-row" + (item.is_flagged ? ' flagged' : '')}>
                 <td style={timeColumn}>
                     <a href={PATH_JIRA_ISSUE + item.key} title={item.summary} target="_blank"
                        rel="noopener noreferrer">{item.key}</a>
--- a/README.rst
+++ b/README.rst
@@ -53,6 +53,11 @@ For each member of the cell this view displays:
 4. Committed time, which is the sum of all estimation hours.
 5. Goal for the next sprint, which is calculated by summing up user's commitments for each day of the sprint and subtracting `SPRINT_HOURS_RESERVED_FOR_MEETINGS` and `Vacation` hours from it.
 6. Remaining time, which is the result of `Goal` - `Committed`.
+7. Toggles:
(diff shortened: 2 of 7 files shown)


shimulch (approved):
LGTM :+1:

- [x] I tested this: Tested locally according to test instruction
- [x] I read through the code
- [x] I checked for accessibility issues
- [x] Includes documentation
- [N/A] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.

frontend/src/components/sprint/UserTable.js
```diff
@@ -31,7 +31,7 @@ const UserTable = ({list, username}) =>
         </thead>
         <tbody>
         {list.map(item =>
-            <tr key={item.key} className={"table-row"}>
+            <tr key={item.key} className={"table-row" + (item.is_flagged ? ' flagged' : '')}>
```
shimulch, line 34: Could be written using template literals which looks much cleaner to me than concat. But serves the same purpose -
```
<tr key={item.key} className={`table-row ${item.is_flagged ? 'flagged' : ''}`}>
```

Agrendalath: @shimulch, TBH nesting braces in one line looks less readable to me in this particular case. I'll leave it as is, because we might revisit it anyway after the UX review.

PR #59, opened 2020-11-15
[BB-3173] Fix calculating vacations and commitments

Agrendalath:
## What it is
This contains the following changes:
1. Fixes calculating commitments and vacations with the assumption that a new sprint starts at midnight UTC.
1. Upgrades Python to `3.9` and Django to `3.1.3` and bumps dependencies (minor changes; unfortunately we cannot use [PEP 585](https://www.python.org/dev/peps/pep-0585/), because mypy [does not support it yet](https://github.com/python/mypy/issues/7907)).
1. Removes unused code from the cookiecutter template.

## Testing instructions 
1. Set `.env` and `frontend/env.local` (you can find them in a comment on the ticket).
1. Comment out [these two lines](https://github.com/open-craft/sprints/blob/master/frontend/src/App.js#L87-L88). They're irrelevant to the scope and will only slow down the local environment.
1. Register new account and get the verification link with `docker-compose -f local.yml logs django`.
1. Log in and navigate to cell's board.
1. Create an "off" calendar event for 4 edge cases described in the `_get_vacation_for_day`'s docstring. Change your timezone [here](https://docs.google.com/spreadsheets/d/107dR9H1vWjLpJlXPuBaFJIFaEPEmaSh50xLow_aEGVw/edit#gid=1116869903) to a negative one to see how it impacts these cases.


sprints/dashboard/models.py
```diff
@@ -377,31 +387,88 @@ def generate_rows(self) -> None:
...
+               the sprint ends. The ratio is represented by `division`.
+        4. For the first day of the future next sprint, there are two subcases:
+            a. Positive timezone - this day can span the next and future next sprint, because user can work after
+               the sprint ends. The ratio is represented by `division`.
+            b. Negative timezone - this day is completely a part of the future next sprint, so this time is ignored (0).
+
+        `division` - a part of the user's availability before the start of the sprint.
+
+        TODO: Check whether this works correctly with other sprint start times than midnight UTC.
+              For these it can span 3 days, so we might have 6 (or even more) corner cases.
```
Agrendalath, line 460: This is not our use case now, so I omitted this.

0x29a (approved):
:+1:

- [x] I tested that all 8 scenarios from `_get_vacation_for_day` docstring work properly for team members from different time zones.
- [x] I read through the code
- [x] Includes documentation

PR #61, opened 2020-11-24
[BB-3173] Fix vacations for negative timezones

Agrendalath:
This fixes a bug with calculating vacations for negative timezones from #59. That PR did not invert the division for negative timezones, so these cases were not being classified correctly:
1. Negative timezone, non-overlapping day, vacations at the last day of the active sprint. These were counted as a part of the next sprint instead of the active one.
2. Negative timezone, non-overlapping day, vacations at the last day of the next sprint. These were counted as a part of the future next sprint instead of the next one.

## Testing instructions:
1. Check that the tests are passing.


0x29a (approved):
:+1:

- [x] I checked that tests are passing.
- [x] I was rereading diff until I understood what caused bug and why change fixes it.
- [x] Includes documentation.

PR #62, opened 2020-11-26
[BB-3176] Added the function to create mattermost post

farhaanbukhsh:
This PR helps to add functionality to create mattermost post, in order to give various information from sprint app to mattermost.

**JIRA tickets**: [BB-3176](https://tasks.opencraft.com/browse/BB-3176)

**Discussions**: [JIRA](https://tasks.opencraft.com/browse/BB-3176?focusedCommentId=176933&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-176933)

**Dependencies**: I have already created a sprint bot on mattermost for the same.

~~**Screenshots**: Always include screenshots if there is any change to the UI.~~

~~**Sandbox URL**: TBD - sandbox is being provisioned.~~

~~**Merge deadline**: "None" if there's no rush, "ASAP" if it's critical, or provide a specific date if there is one.`~~

**Testing instructions**:

* In the .env file we need to put values
```
# MATTERMOST
# ------------------------------------------------------------------------------
# URL of the Mattermost server.
MATTERMOST_SERVER=chat.opencraft.com
# Default is the https port but it can be configured
MATTERMOST_PORT=443
# Channel to ping
MATTERMOST_CHANNEL=botbattleground
# Login Id
MATTERMOST_LOGIN_ID=sprintbot
# Access Token
MATTERMOST_ACCESS_TOKEN=<token>
# Team Name
MATTERMOST_TEAM_NAME=OpenCraft
```

1. I have attached the vault link for the same file in the ticket
1. Checkout to this branch
1. Run `docker-compose -f local.yml up --build`
1. Get the shell excess `docker exec -it sprints_django_1 /bin/sh`
1. `export DJANGO_SETTINGS_MODULE=config.settings.local`
1. In the shell run `/app # ipython3`
1. 
```
In [4]: from sprints.dashboard.utils import create_mattermost_post

In [5]: create_mattermost_post('This is test from method', ['farhaan@opencraft.com'])
```
* you will see the message got posted on the `botbattleground` channel

**Author notes and concerns**:

I created a dummy channel so that testing should not bother anyone.

**Reviewers**
- [ ] @Agrendalath


sprints/dashboard/libs/mattermost.py
```diff
@@ -0,0 +1,70 @@
...
+        username = self.mattermost_connection.users.get_user_by_email(email=email_id)['username']
+        return username
+
+    def post_message_to_channel(self, channel_name: str, message: str) -> None:
+        """
+        Post the message to the channel using the channel name.
+
+        Args:
+            channel_name: Name of the channel on mattermost.
+            message: Message to be posted on the channel.
```
0x29a, line 47: Nit: check out other docstrings in this repo. They're in `reStructuredText` format, while this is `Google Style`.

sprints/dashboard/utils.py
```diff
@@ -1,4 +1,5 @@
 import re
+from sprints.dashboard.libs.mattermost import connect_to_mattermost
```
0x29a, line 2: Nit: can you place this import under STL import statements?

Agrendalath (changes requested):
@farhaanbukhsh, looks neat! I left few comments about handling the exceptions and some nits.

sprints/dashboard/utils.py
```diff
@@ -605,3 +606,20 @@ def get_sprint_meeting_day_division(sprint_start: str) -> DefaultDict[str, Tuple
             break
 
     return result
+
+
+def create_mattermost_post(message: str, email_ids: List[str]) -> None:
```
Agrendalath, line 611: I think it might make sense to move this to `libs.mattermost`. I took a second look at this file and the number of lines grew tremendously here since I last checked it.

Agrendalath:
@farhaanbukhsh, also a small remark about the commit message. Please use the imperative mood to produce more concise message and then add a more specific description below in the message, i.e.
```
Add Mattermost integration

This adds a wrapper for Mattermost driver to easily send messages that let people know about urgent events.
```
Or something like this.

sprints/dashboard/utils.py
farhaanbukhsh: I agree, also files often like `common` and `utils` tends to become a dumping ground for functions, but maybe that refactoring is a part of another day. ;)

Agrendalath: Yep, let's leave the general refactor for another time, but we can avoid creating more functions here.

farhaanbukhsh:
> @farhaanbukhsh, also a small remark about the commit message. Please use the imperative mood to produce more concise first line of the message and then add a more specific description below, i.e.
> 
> ```
> Add Mattermost integration
> 
> This adds a wrapper for Mattermost driver to easily send messages that let people know about urgent events.
> ```
> 
> Or something like this.

Makes sense, I should have been more careful with this, thanks for pointing out :)

sprints/dashboard/libs/mattermost.py
```diff
@@ -0,0 +1,67 @@
...
+        :param username: Username of the mattermost bot.
+        :type username: str
+        :param access_token: Access token for the bot.
+        :type access_token: str
+        """
+        connection_parameter = {
+            'url': settings.MATTERMOST_SERVER,
+            'port': settings.MATTERMOST_PORT,
+            'username': username,
+            'token': access_token
```
Agrendalath, line 21: @farhaanbukhsh, ok, my last comment and I'm going to sleep :P
Could we use the `settings.MATTERMOST_LOGIN_ID, settings.MATTERMOST_ACCESS_TOKEN` here and make this class a context manager itself (by implementing `__enter__` and `__exit__`) instead of adding `connect_to_mattermost`? I know that you took it from other libs, but I think that in this case it might be a better approach, as we're creating a new class that encapsulates the Mattermost driver. What do you think?

farhaanbukhsh: Actually I did think about it :) and the reason I went with the current approach was because I didn't want to break the pattern. For example all lib files are following a pattern so I thought why not keep it consistent.

farhaanbukhsh: and I am happy to do this by implementing `__enter__` and `__exit__` 👍

farhaanbukhsh: also I don't want to use `settings.MATTERMOST_LOGIN_ID, settings.MATTERMOST_ACCESS_TOKEN` because what if we will use multiple bots?

Agrendalath: > what if we will use multiple bots?

Let's have only one bot for now; finding cool names for more of them is going to take a lot of time ;)
Jokes aside, the current code would need some changes to support more of them anyway, e.g. specifying different `username:token` pairs in the configurations, a way to pass the correct pair to the context manager, etc. So let's not worry about this, as we don't need it now.

farhaanbukhsh: sure thing 👍

sprints/dashboard/libs/mattermost.py
```diff
@@ -0,0 +1,89 @@
...
+            from sentry_sdk import capture_exception
+            capture_exception(e)
+        return self
+
+    def __exit__(self, *args):
+        self.mattermost_connection.logout()
+
+    def get_usernames_from_emails(self, emails: List[str]) -> List[str]:
+        """
+        Function that helps to get mattermost usrenames
```
Agrendalath, line 38: Typo: `usernames`.

sprints/dashboard/libs/mattermost.py
```diff
@@ -0,0 +1,89 @@
...
+        from emails.
+
+        :param emails: Emails of the users.
+        :return: Mattermost usernames of the users.
+        """
+        usernames = []
+        for email in emails:
+            try:
+                username = self.mattermost_connection.users.get_user_by_email(email=email).get('username')
```
Agrendalath, line 47: Nit: `email=email` is redundant, as there is only one possible argument for this method.

sprints/dashboard/libs/mattermost.py
```diff
@@ -0,0 +1,89 @@
...
+                if username:
```
Agrendalath, line 48: When can this value be non-truthy? Doesn't it just raise 404 when user has not been found?

sprints/dashboard/tests/test_libs.py
```diff
@@ -0,0 +1,47 @@
+from unittest.mock import patch
```
Agrendalath, line 1: We could place this code in `sprints/dashboard/libs/tests/test_mattermost.py`. What do you think?

sprints/dashboard/tests/test_libs.py
```diff
@@ -0,0 +1,47 @@
...
+
+from django.test import override_settings
+
+from sprints.dashboard.libs.mattermost import create_mattermost_post
+
+
+def side_effect_usernames_from_emails(emails):
+    usernames = []
+    username_email_map = {'farhaan@example.com': 'farhaan',
+                          'piotr@example.com': 'piotr', 'demid@example.com': 'demid'}
```
Agrendalath, line 11: Nit: let's make this a one-liner. This formatting looks a bit odd.

sprints/dashboard/libs/mattermost.py
```diff
@@ -0,0 +1,89 @@
...
+        }
+        self.mattermost_connection = Driver(connection_parameters)
+
+    def __enter__(self):
+        try:
+            self.mattermost_connection.login()
+        except HTTPError as e:
+            # noinspection PyUnresolvedReferences
+            from sentry_sdk import capture_exception
+            capture_exception(e)
```
Agrendalath, line 30: You skipped the `if not settings.DEBUG:` check. Please see [this](https://github.com/open-craft/sprints/blob/master/sprints/dashboard/utils.py#L556-L562) again. There is no point in using Sentry locally.

sprints/dashboard/libs/mattermost.py
```diff
@@ -0,0 +1,89 @@
...
+        usernames = []
+        for email in emails:
+            try:
+                username = self.mattermost_connection.users.get_user_by_email(email=email).get('username')
+                if username:
+                    usernames.append(username)
+            except HTTPError as e:
+                # noinspection PyUnresolvedReferences
+                from sentry_sdk import capture_exception
+                capture_exception(e)
```
Agrendalath, line 53: You skipped the `if not settings.DEBUG:` check. Please see [this](https://github.com/open-craft/sprints/blob/master/sprints/dashboard/utils.py#L556-L562) again. There is no point in using Sentry locally.

sprints/dashboard/libs/mattermost.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        try:
+            channels = self.mattermost_connection.channels
+            channel_id = channels.get_channel_by_name_and_team_name(
+                settings.MATTERMOST_TEAM_NAME, channel_name).get('id')
+            self.mattermost_connection.posts.create_post({'channel_id': channel_id, 'message': message})
+        except HTTPError as e:
+            # noinspection PyUnresolvedReferences
+            from sentry_sdk import capture_exception
+            capture_exception(e)
```
Agrendalath, line 72: You skipped the `if not settings.DEBUG:` check. Please see [this](https://github.com/open-craft/sprints/blob/master/sprints/dashboard/utils.py#L556-L562) again. There is no point in using Sentry locally.

config/settings/base.py
```diff
@@ -611,3 +611,18 @@
...
 MULTIPROCESSING_TIMEOUT = env.int("MULTIPROCESSING_TIMEOUT", 128)
+
+# MATTERMOST
+# ------------------------------------------------------------------------------
+# URL of the Mattermost server.
+MATTERMOST_SERVER = env.str("MATTERMOST_SERVER")
+# Default is the https port but it can be configured
+MATTERMOST_PORT = env.int("MATTERMOST_PORT", 443)
+# Channel to ping
+MATTERMOST_CHANNEL = env.str("MATTERMOST_CHANNEL")
```
Agrendalath, line 622: Let's add a note (both here and in local and production `.envs/`) that the channel name should be lowercase and that the spaces should be replaced with minus signs (e.g. "Sprint Planning" -> "sprint-planning").

Agrendalath (approved):
@farhaanbukhsh, leaving the conditional approval. Please address all comments and then use the `Squash & Merge` feature.

Also two remarks for the future:
- Please don't squash commits and don't make force pushes before finishing the review. It is impossible to see the subsets of changes between versions then.
- Please add some indicator after addressing the comments, so the reviewer knows that it has been addressed - a +1 "reaction" under each comment should be sufficient.

:+1:
* [x]  I tested this: tested that the Mattermost integration is working
* [ ]  Includes documentation: no :warning: I can add it in BB-3175, so it won't block this.

PR #63, opened 2020-12-29
[BB-3452] Adding ‘Start of New Sprint’ webhook dispatcher to Sprints

aayushagra:
Functionality to define member roles (Sprint Planning Manager etc) for the upcoming crafty checklisting feature

**Merge deadline**: None

**Testing instructions**:

1. Checkout this branch
2. Run migrations (docker-compose run django python manage.py migrate)
3. Open the admin panel (localhost:8000/admin)
4. Create a new webhook event called 'new sprint'
5. Create a new webhook and set the webhook url to a temporary request receiver (eg: https://requestbin.com/), link to the webhook event created in the last step
6. Set the HANDBOOK_ROLES_PAGE env variable to 'https://handbook.opencraft.com/en/latest/cells/'
7. Set the FEATURE_CELL_ROLES env variable to True
8. End sprints for each cell and verify that the Webhook receiver receives a list of that cell's members and a list of their roles. Firefighter & Discovery duty should also be in the list as FF-1, FF-2, DD-1, DD-2

**Reviewers**
- [x] @Agrendalath


aayushagra:
@Agrendalath The tests are failing on CircleCI because an env var is missing, but I'm not sure where to define it for CircleCI since a lot of other variables have been defined in the same way and don't seem to be defined elsewhere. Do you know where I should define these new env vars for CircleCI?

config/settings/base.py
```diff
@@ -629,3 +629,12 @@
 MATTERMOST_ACCESS_TOKEN = env.str("MATTERMOST_ACCESS_TOKEN")
 # Team Name
 MATTERMOST_TEAM_NAME = env.str("MATTERMOST_TEAM_NAME")
+
+# WEBHOOKS
+# ------------------------------------------------------------------------------
+# Webhook username (Used with basic auth)
+WEBHOOK_USERNAME = env.str("WEBHOOK_USERNAME")
+# Webhook password (Used with basic auth)
+WEBHOOK_PASSWORD = env.str("WEBHOOK_PASSWORD")
```
gabor-boros, line 638: In case the `WEBHOOK_USERNAME` and `WEBHOOK_PASSWORD` indicates access to the crafty bot application, then it should be a `WEBHOOK_TOKEN` (or `CRAFTY_BOT_TOKEN` may be better) and the requested should have `headers={"Authorization": f"Token {CRAFTY_BOT_TOKEN}"}` 😇

Agrendalath:
> The tests are failing on CircleCI because an env var is missing, but I'm not sure where to define it for CircleCI since a lot of other variables have been defined in the same way and don't seem to be defined elsewhere. Do you know where I should define these new env vars for CircleCI?

@aayushagra, they could be defined in [pytest.ini](https://github.com/open-craft/sprints/blob/master/pytest.ini), **but** I believe they shouldn't be there. The webhooks should be considered as an optional feature, so instead of forcing these variables to be defined, we should provide default `None` values for them in `config/settings/base.py` and avoid triggering webhooks in case these variables are not defined.

sprints/dashboard/views.py
```diff
@@ -96,6 +98,11 @@ def can_end_sprint(board_id: int, acquire_lock: bool = False) -> Tuple[bool, str
         if datetime.now() < end_date:
             return False, "The current day is not the last day of the current sprint."
 
+        try:
+            get_cell_member_roles(raise_exception=True)
+        except InsufficientRolesReadError:
+            return False, "Unable to read roles from the handbook."
```
Agrendalath, line 104: If we're catching an exception here, shouldn't we pass its message as an error?

sprints/dashboard/utils.py
```diff
@@ -97,6 +100,37 @@ def get_cell_member_names(conn: CustomJira, members: Iterable[str]) -> Dict[str,
...
+    roles_dict = {}
+    for role, member in roles:
+        if member not in roles_dict:
+            roles_dict[member] = [role]
+        else:
+            roles_dict[member].append(role)
+
+    # It's difficult to know whether we have read all roles
+    # So we assume that if we've read roles for less than 12 members, something's probably wrong
+    if raise_exception and len(roles_dict) < 12:
```
Agrendalath, line 129: This does not look right. Maybe let's just check if it's not empty instead of hardcoding this?
cc: @nizarmah

sprints/dashboard/utils.py
```diff
@@ -97,6 +100,37 @@ def get_cell_member_names(conn: CustomJira, members: Iterable[str]) -> Dict[str,
...
+    # roles = [('Recruitment manager', 'John Doe'),('Sprint Planning Manager', 'John Doe'),...]
+    roles = re.findall(ROLES_REGEX, r.text)
+
+    # roles_dict = {'John Doe': ['Recruitment Manager', 'Sprint Planning Manager',...],...}
+    roles_dict = {}
+    for role, member in roles:
+        if member not in roles_dict:
+            roles_dict[member] = [role]
+        else:
+            roles_dict[member].append(role)
```
Agrendalath, line 125: Couldn't we just use `defaultdict(list)`?

sprints/dashboard/tasks.py
```diff
@@ -202,11 +265,21 @@ def complete_sprint_task(board_id: int) -> None:
...
+        }
+
         # It is not mentioned in Python lib docs, but the limit for the issue-moving queries is 50 issues. Source:
         # https://developer.atlassian.com/cloud/jira/software/rest/#api-rest-agile-1-0-backlog-issue-post
         # https://developer.atlassian.com/cloud/jira/software/rest/#api-rest-agile-1-0-sprint-sprintId-issue-post
         batch_size = 50
         if not settings.DEBUG:  # We really don't want to trigger this in the dev environment.
+            # Raise error if we can't read roles from the handbook
+            get_cell_member_roles(raise_exception=True)
+            trigger_new_sprint_webhooks.delay(cell_dict, next_sprint.name, get_sprint_number(next_sprint), board_id)
```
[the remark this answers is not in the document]
nizarmah: > Shouldn't we trigger this after ending the sprint?

I agree, the event should only be sent after ending the sprint, and the new sprint has already started.

> Will the receiving part have some safety checks that will verify the sprint number? 

The receiving party, for now, has no safety checks that will verify the sprint number.
We expect the receiving party to use the information without "second guessing" it.

sprints/dashboard/utils.py
nizarmah: > Maybe let's just check if it's not empty instead of hardcoding this?

Yeah I support just checking if it's empty instead of hard coding such a condition 👍🏼

sprints/dashboard/tasks.py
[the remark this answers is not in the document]
gabor-boros: Sorry for sneaking in. @nizarmah I would say that the receiving party should use the information without second-guessing the incoming value **but** it **should** make sure that the sprint id is unique and not 0.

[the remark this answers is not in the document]
nizarmah: No worries @gabor-boros 

> it should make sure that the sprint id is unique and not 0

But if the sprint id is 0 or not unique, the receiving party would simply not do anything about that. Instead, it would dismiss such a received event; please correct me if I'm wrong.

Based on that, there shouldn't be an issue/problem with trigger the event after ending the sprint.

---

Also the `get_cell_member_roles` and the `trigger_new_sprint_webhooks` don't have to be placed under the same condition statement. 

We want them to actually be placed in two different places in the method, so it would be better to separate them from under that conditional.

We do want to run the `get_cell_member_roles` though before ending the sprint, because we don't want to interrupt ending the sprint if we don't have any roles; instead, it would be better not to end the sprint at all, until we have the roles.

But the triggering of the event, that needs to be done at the end, after all the code that would be responsible for ending the sprint. And in case that's asynchronous, it should be done using a callback.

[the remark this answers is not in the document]
gabor-boros: @nizarmah From API point of view it would be expected to return 400 when an invalid data received, so - in my opinion - client should handle that scenario (and not crash).

One way of that could be to simply send an email to us about the fact the request failed.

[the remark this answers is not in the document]
nizarmah: @gabor-boros but this should be a webhook event. So, any triggered/sent event should be a valid event, or the only source of truth. We're not waiting for a response, we're just posting data...

Crafty, a third party, should not be aware of what a valid Sprints "new sprint" event is... Crafty should only be aware of whether an event is useful for it or not. For example, let's say Crafty receives an event with no roles assigned to the members... Crafty should treat that event as the only source of truth and assign to the different members only tasks that are for Cell Members.

Now, sure, we might choose to add a condition so that Crafty doesn't create new checklists when the roles are empty, but that doesn't make the Sprints event any less valid in the eyes of Crafty, it just makes it inappropriate for its use.

So, for example, simply we'd have a condition that checks that the sprint id in the received event is valid. Otherwise, we'd **have Crafty ping** the _urgent_ mailing list.

If an invalid webhook event is triggered, then we haven't verified the data before sending an event. Any validation needs to be done in Sprints, and not in Crafty. If we return 400 and Sprints actually listens to the response, that means we haven't separated the services from each other, which is the main goal behind having Crafty.

And yes, I know that this is being said, while I'm requesting that we rely on the handbook to read the roles. And I know that we will use Crafty to create different checklists on monday.com. 😔 

But once we implement the open source solution, I have a bunch of work planned for Crafty, to get it closer to being a service (until we have an epic for it).

sprints/dashboard/tasks.py
```diff
@@ -144,19 +148,78 @@ def create_role_issues_task(cell: Dict[str, str], sprint_id: int, sprint_number:
...
     1. Uploads spillovers.
     2. Uploads commitments.
-    3. Moves archived issues out of the active sprint.
-    4. Closes the active sprint.
-    5. Moves issues from the closed sprint to the next one.
-    6. Opens the next sprint.
-    7. Creates role tickets.
-    8. Releases the sprint completion lock.
-    9. Clears cache related to end of sprint date.
+    3. Trigger start of new sprint webhooks
```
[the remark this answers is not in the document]
aayushagra: @Agrendalath Can you please clarify what you mean?

sprints/dashboard/tasks.py
```diff
@@ -144,19 +148,78 @@ def create_role_issues_task(cell: Dict[str, str], sprint_id: int, sprint_number:
...
                     conn.create_issue(fields)
 
+@celery_app.task(ignore_result=True)
+def trigger_new_sprint_webhooks(cell: Dict[str, str], sprint_name: str, sprint_number: int, board_id: int):
+    """
+    1. Gather a dictionary mapping Name to E-Mail Address of members
+    2. Gather cell member roles
+    3. Map E-Mails to Cell member roles
+    4. Add rotations (DD, FF etc) to the Cell member roles array
+    5. Trigger webhooks
```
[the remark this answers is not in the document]
aayushagra: @Agrendalath Can you please clarify what you mean?

sprints/dashboard/tasks.py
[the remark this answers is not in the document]
Agrendalath: @aayushagra, the mood you used is imperative ("Trigger"), but the rest of this docstring is indicative ("Creates", "Releases", "Clears"), so it should be "Triggers" to match the other lines.

sprints/dashboard/tasks.py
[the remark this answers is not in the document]
Agrendalath: @aayushagra, please see [this answer](https://github.com/open-craft/sprints/pull/63#discussion_r552197712). We might not have been consistent about this in the whole project, but we could make these moods match within one file for now.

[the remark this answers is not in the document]
aayushagra: @Agrendalath In that case the rest of the docstring for the function is still using the imperative mood. Would you like me to just go ahead and set the entire docstring to the indicative mood?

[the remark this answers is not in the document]
Agrendalath: @aayushagra, I mentioned keeping the consistency within a file for this case, but I'll leave the decision to you here :)

sprints/dashboard/utils.py
```diff
@@ -97,6 +101,67 @@ def get_cell_member_names(conn: CustomJira, members: Iterable[str]) -> Dict[str,
     """Returns cell members with their names."""
     return {conn.user(member).displayName: member for member in members}
```
Agrendalath, line 103: There should be two lines of space between these functions. Not sure why `pylint` didn't catch this…
Edit. ah, I see that we had to disable it because of some compatibility problems. Therefore please be extra careful about PEP-8…

sprints/dashboard/utils.py
```diff
@@ -97,6 +101,67 @@ def get_cell_member_names(conn: CustomJira, members: Iterable[str]) -> Dict[str,
...
+    if raise_exception and len(roles_dict) == 0:
+        raise NoRolesFoundException("No roles were found at the handbook page: %s" % HANDBOOK_ROLES_PAGE)
+
+    return roles_dict
+
+def compile_participants_roles(
+    members: List[User],
+    rotations: Dict[str, List[str]],
+    cell_member_roles: Dict[str, List[str]]
+) -> Dict[str, List[str]]:
```
Agrendalath, line 136: It's this `DefaultDict`, as the return value and in this function's args? These are crucial for this function's logic, so we should be strict about it.

sprints/dashboard/utils.py
```diff
@@ -97,6 +101,67 @@ def get_cell_member_names(conn: CustomJira, members: Iterable[str]) -> Dict[str,
...
+            cell_member_roles[member.displayName]
+        )
+
+        roles[member.emailAddress].extend(
+            get_rotations_roles_for_member(member.displayName, rotations)
+        )
+
+    return roles
+
+def get_rotations_roles_for_member(member_name: str, rotations: Dict[str, List[str]]):
```
Agrendalath, line 152: Nit: please add the return type.

sprints/dashboard/utils.py
```diff
@@ -97,6 +101,67 @@ def get_cell_member_names(conn: CustomJira, members: Iterable[str]) -> Dict[str,
     """Returns cell members with their names."""
     return {conn.user(member).displayName: member for member in members}
 
+def get_cell_member_roles(raise_exception: bool = False) -> Dict[str, List[str]]:
+    """Return a dictionary of cell members and their associated roles.
+    Example return: {'John Doe': ['Recruitment Manager', 'Sprint Planning Manager',...],...}.
+
+    Note: We have room for error here if member names are spelled differently, hence we must
+    be careful that the handbook & Jira have consistent naming.
```
Agrendalath, line 109: Maybe something along this would be more verbose?
```
Note: we are using the following sources of data for this function:
  1. Handbook (cell roles, e.g. "Sprint Planning Manager").
  2. Rotations spreadsheet (sprint roles - e.g. "Firefighter", "Discovery Duty")
  3. Jira (matching users with their email addresses).
Therefore we should ensure that users' names don't contain any typos, as this will provide inaccurate results.
```

sprints/dashboard/utils.py
```diff
@@ -97,6 +101,67 @@ def get_cell_member_names(conn: CustomJira, members: Iterable[str]) -> Dict[str,
...
+
+    # roles_dict = {'John Doe': ['Recruitment Manager', 'Sprint Planning Manager',...],...}
+    roles_dict = defaultdict(list)
+    for role, member in roles:
+        roles_dict[member].append(role)
+
+    # It's difficult to know whether we have read all roles
+    # So we assume that if we've read roles for less than 12 members, something's probably wrong
+    if raise_exception and len(roles_dict) == 0:
+        raise NoRolesFoundException("No roles were found at the handbook page: %s" % HANDBOOK_ROLES_PAGE)
```
Agrendalath, line 128: Any reason not to use f-strings here? They are much more concise and elegant.

sprints/dashboard/utils.py
```diff
@@ -97,6 +101,67 @@ def get_cell_member_names(conn: CustomJira, members: Iterable[str]) -> Dict[str,
...
+    """Given a member and a dictionary containing get_rotations_users() output, return a list
+    of all roles for that user.
+    """
+
+    roles = []
+    for duty, assignees in rotations.items():
+        for idx, assignee in enumerate(assignees):
+            # The rotations sheet sometimes contains only the first name
+            if member_name.startswith(assignee):
+                roles.append("%s-%d" % (duty, idx+1))
```
Agrendalath, line 162: Please use f-strings, they are much more readable.

sprints/dashboard/utils.py
```diff
@@ -97,6 +101,67 @@ def get_cell_member_names(conn: CustomJira, members: Iterable[str]) -> Dict[str,
...
+
+def get_rotations_roles_for_member(member_name: str, rotations: Dict[str, List[str]]):
+    """Given a member and a dictionary containing get_rotations_users() output, return a list
+    of all roles for that user.
+    """
+
+    roles = []
+    for duty, assignees in rotations.items():
+        for idx, assignee in enumerate(assignees):
+            # The rotations sheet sometimes contains only the first name
```
Agrendalath, line 160: Typo: `sometimes contains` -> `might contain`.

sprints/dashboard/utils.py
Agrendalath: Also what's the point of this enumeration? It's confusing; could we make this clear in the docstring?

sprints/dashboard/tests/test_utils.py
```diff
@@ -322,3 +327,70 @@ def test_column_number_to_excel(test_input, expected):
...
+def test_get_cell_member_roles_corrupted(mock_get):
+    with open("sprints/dashboard/tests/data/handbook/dummy_cells_corrupted.html", 'r') as f:
+        mock_get.return_value = Mock(text=f.read())
+        with TestCase().assertRaises(NoRolesFoundException):
+            get_cell_member_roles(raise_exception=True)
+
+def test_get_rotations_roles_for_member():
+    # When the member is on FF duty
+    output = get_rotations_roles_for_member('John Doe', {'FF': ['Jake Doe', 'John Doe'], 'DD': ['Jane Doe', 'James Doe']})
+    assert len(output) == 1 and output[0] == 'FF-2'
```
Agrendalath, line 357: Please separate the asserts. They should be single-purposed to have an immediate view of what is failing.

sprints/dashboard/tests/test_utils.py
```diff
@@ -322,3 +327,70 @@ def test_column_number_to_excel(test_input, expected):
...
+        output = dict(get_cell_member_roles())
+        expected_output = {
+                            "Member Six": ["Recruitment manager"],
+                            "Member Seven": ["Sprint manager"],
+                            "Member Five": ["Sprint Planning Manager"],
+                            "Member Eight": ["Epic manager"],
+                            "Member Two": ["Sustainability manager"],
+                            "Member Four": ["OSPR Liaison", "Official forum moderator"],
+                            "Member One": ["DevOps Specialist"],
+                        }
```
Agrendalath, line 344: This formatting is very odd. Please run `black` on all changes, as suggested in another comment.

sprints/dashboard/tests/test_utils.py
```diff
@@ -322,3 +327,70 @@ def test_column_number_to_excel(test_input, expected):
...
+    output = compile_participants_roles(members_data_dummy, rotations_data_dummy, roles_data_dummy)
+
+    expected_output = {
+        'janedoe@opencraft.com': [],
+        'jackdoe@opencraft.com': ['Sprint Planning Manager', 'DevOps Specialist', 'ff-2', 'dd-1'],
+        'johndoe@opencraft.com': ['Sprint Manager', 'ff-1'],
+        'jakedoe@opencraft.com': ['Recruitment Manager', 'dd-2']
+    }
+
+    TestCase().assertDictEqual(output, expected_output)
```
Agrendalath, line 396: It might make sense to verify case with empty data too, to be sure that all corner cases will remain handled if we make any changes to these functions.

sprints/webhooks/models.py
```diff
@@ -0,0 +1,25 @@
+from django.db import models
+import requests
+import json
```
Agrendalath, line 3: Please sort imports as [defined is PEP-8](https://www.python.org/dev/peps/pep-0008/#imports):
```python
import json

import requests
from django.db import models
```

sprints/dashboard/tasks.py
```diff
@@ -3,9 +3,11 @@
     datetime,
     timedelta,
 )
+from collections import defaultdict
```
Agrendalath, line 6: Are we using this import?

sprints/dashboard/tasks.py
```diff
@@ -3,9 +3,11 @@
...
 from typing import (
     Dict,
     List,
+    DefaultDict,
```
Agrendalath, line 10: Are we using this import?

sprints/dashboard/tasks.py
```diff
@@ -28,11 +30,14 @@
 )
 from sprints.dashboard.libs.jira import connect_to_jira
 from sprints.dashboard.models import Dashboard
+from sprints.webhooks.models import Webhook, WebhookEvent
```
Agrendalath, line 33: This should be below `sprints.dashboard.utils` (it's sorted alphabetically). Also `WebhookEvent` seems to be unused.

sprints/dashboard/tasks.py
```diff
@@ -28,11 +30,14 @@
...
 from sprints.dashboard.utils import (
     create_next_sprint,
     filter_sprints_by_cell,
     get_all_sprints,
     get_cell_member_names,
+    get_cell_member_roles,
+    compile_participants_roles,
```
Agrendalath, line 40: Nit: this one should be at the top of this import.

sprints/dashboard/tasks.py
```diff
@@ -144,6 +149,44 @@ def create_role_issues_task(cell: Dict[str, str], sprint_id: int, sprint_number:
...
+    """
+    1. Gathers a dictionary mapping Name to E-Mail Address of members
+    2. Gathers cell member roles
+    3. Maps cell member's E-Mails to their roles
+    4. Adds rotations (DD, FF etc) to the cell member roles array
+    5. Triggers webhooks
+
+    The webhook receivers identify users by E-Mail, but our spreadsheets & documentation identify them
+    by name, so we must map the two. It is important that the spelling of member names is consistent in
+    the documentation.
```
Agrendalath, line 163: Please see the comment in `utils.py` about this. Let's not duplicate docstrings, as it's hard to keep them in sync then. Besides I think that this function does not do the mapping by itself, right?

Agrendalath (changes requested):

sprints/dashboard/tasks.py
```diff
@@ -144,6 +149,44 @@ def create_role_issues_task(cell: Dict[str, str], sprint_id: int, sprint_number:
...
                     conn.create_issue(fields)
 
+@celery_app.task(ignore_result=True)
+def trigger_new_sprint_webhooks(cell_name: str, sprint_name: str, sprint_number: int, board_id: int):
+    """
+    1. Gathers a dictionary mapping Name to E-Mail Address of members
+    2. Gathers cell member roles
+    3. Maps cell member's E-Mails to their roles
+    4. Adds rotations (DD, FF etc) to the cell member roles array
+    5. Triggers webhooks
```
Agrendalath, line 159: Is it still accurate?

sprints/dashboard/tasks.py
```diff
@@ -144,6 +149,44 @@ def create_role_issues_task(cell: Dict[str, str], sprint_id: int, sprint_number:
 
                     conn.create_issue(fields)
 
+@celery_app.task(ignore_result=True)
+def trigger_new_sprint_webhooks(cell_name: str, sprint_name: str, sprint_number: int, board_id: int):
```
Agrendalath, line 153: Please add the `_task` prefix to maintain consistency with other tasks.

sprints/dashboard/tasks.py
```diff
@@ -245,13 +299,10 @@ def complete_sprint_task(board_id: int) -> None:
                 future_next_sprint_number = create_next_sprint_task(board_id)
                 future_next_sprint = get_next_sprint(sprints, next_sprint)
 
-            cell_dict = {
-                'key': cell.key,
-                'name': cell.name,
-                'board_id': cell.board_id,
-            }
```
Agrendalath, line 252: I guess we don't need to change this anymore, right?

sprints/webhooks/models.py
```diff
@@ -0,0 +1,25 @@
...
+
+class WebhookEvent(models.Model):
+    name = models.TextField()
+
+    def __str__(self):
+        return self.name
+
+class Webhook(models.Model):
+    payload_url = models.URLField()
+    active = models.BooleanField()
```
Agrendalath, line 13: Where are we using this?

sprints/webhooks/models.py
```diff
@@ -0,0 +1,25 @@
+from django.db import models
+import requests
+import json
+
+class WebhookEvent(models.Model):
+    name = models.TextField()
```
Agrendalath, line 6: One last comment, as I've just noticed this wile testing the PR - why aren't we simply using `CharField` here? We don't need textarea in the admin panel for this, as it's just a name.

Agrendalath:
@aayushagra, I tested this and it works well (except the part with unspecified handbook URL that I mentioned in the review).

sprints/dashboard/utils.py
```diff
@@ -97,6 +101,67 @@ def get_cell_member_names(conn: CustomJira, members: Iterable[str]) -> Dict[str,
...
+    Example return: {'John Doe': ['Recruitment Manager', 'Sprint Planning Manager',...],...}.
+
+    Note: We have room for error here if member names are spelled differently, hence we must
+    be careful that the handbook & Jira have consistent naming.
+    """
+
+    # Example HTML: <li><a href="../roles/#cell-manager-recruitment">Recruitment manager</a>: John Doe</li>
+    ROLES_REGEX = r"<li>.*roles.*>([A-Za-z ]+).*: (.+)<\/li>"
+
+    r = requests.get(HANDBOOK_ROLES_PAGE)
```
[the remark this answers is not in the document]
nizarmah: Addressed 👍🏼

sprints/dashboard/utils.py
```diff
@@ -97,6 +101,67 @@ def get_cell_member_names(conn: CustomJira, members: Iterable[str]) -> Dict[str,
...
+    # roles = [('Recruitment manager', 'John Doe'),('Sprint Planning Manager', 'John Doe'),...]
+    roles = re.findall(ROLES_REGEX, r.text)
+
+    # roles_dict = {'John Doe': ['Recruitment Manager', 'Sprint Planning Manager',...],...}
+    roles_dict = defaultdict(list)
+    for role, member in roles:
+        roles_dict[member].append(role)
+
+    # It's difficult to know whether we have read all roles
+    # So we assume that if we've read roles for less than 12 members, something's probably wrong
```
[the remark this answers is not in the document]
nizarmah: Addressed 👍🏼

sprints/dashboard/utils.py
nizarmah: Addressed 👍🏼

nizarmah (commented):
I created a pull request to address some of the comments raised in this pull request. 

https://github.com/open-craft/sprints/pull/66

I'm pretty sure I broke some tests, so I'll need to look into what broke them exactly later today.

sprints/dashboard/tasks.py
Agrendalath: @nizarmah, this was an unnecessary change, which was confusing, as this variable was not used anywhere earlier. Reverted in d916240.

sprints/dashboard/utils.py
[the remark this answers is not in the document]
Agrendalath: Almost - fixed in 8c8c8ef.

sprints/webhooks/models.py
Agrendalath: @nizarmah, almost - fe4038e.
When updating models, we also need to add/update migrations. Also, it's a good practice to add `max_length` to the [CharField](https://docs.djangoproject.com/en/3.1/ref/models/fields/#django.db.models.CharField).

sprints/dashboard/utils.py
Agrendalath: 39ab8d5

sprints/dashboard/utils.py
Agrendalath: 39ab8d5

Agrendalath: Note: this is not the case anymore, as we've updated the page yesterday and now it contains full names.

sprints/dashboard/tests/test_utils.py
Agrendalath: 39ab8d5

sprints/dashboard/tests/test_utils.py
Agrendalath: @nizarmah, yep, that's what I meant :+1:

sprints/dashboard/utils.py
Agrendalath: 17905db8

sprints/dashboard/utils.py
Agrendalath: 4690c3ee

sprints/dashboard/utils.py
```diff
@@ -97,6 +101,67 @@ def get_cell_member_names(conn: CustomJira, members: Iterable[str]) -> Dict[str,
...
+
+    roles: DefaultDict[str, List[str]] = defaultdict(list)
+    for member in members:
+        roles[member.emailAddress].extend(
+            cell_member_roles[member.displayName]
+        )
+
+        roles[member.emailAddress].extend(
+            get_rotations_roles_for_member(member.displayName, rotations)
+        )
```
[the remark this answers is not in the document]
Agrendalath: 39ab8d5c - partially.

Agrendalath (approved):
:+1:
* [x]  I tested this: tested that webhooks work and the roles are present
* [x]  Includes documentation

sprints/webhooks/models.py
nizarmah: Ah yeah, I keep forgetting about migrations 🤦🏼 

Thanks for the note about adding `max_length` for the CharField 👍🏼

PR #64, opened 2021-01-15
[BB-3566] Fix name collisions and Google auth

Agrendalath:
This:
1. Fixes Google auth, which broke due to two reasons:
1.1. The ones described in open-craft/django-rest-auth/pull/1.
1.2. Removal of required views as a part of the cleanup in #59.
1. Adds support for full names in vacation calendar events.

## Testing instructions 
1. Set `.env` and `frontend/env.local` (you can find them in a comment in the Vault).
1. Comment out [these two lines](https://github.com/open-craft/sprints/blob/master/frontend/src/App.js#L87-L88). They're irrelevant to the scope and will only slow down the local environment.
1. Start the backend with `docker-compose -f local.yml up --build` (this is important because we're changing dependencies here).
1. Log in via Google Auth.
1. Go to [Bebop board](http://localhost:3000/board/26) and check that Daniel does not have any vacations.
1. Go to [Serenity board](http://localhost:3000/board/24) and check that Daniel has 10h of vacation.
1. [Optional] Create some temporary vacation events for both Daniels on the Bebop calendar and check that they work as expected. **Remember to delete them later.**

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -538,10 +538,11 @@ GOOGLE_API_CREDENTIALS = {
 }
 # Regex for retrieving users' vacations from Google Calendar. This one is case-insensitive.
 # It supports two basic cases:
-# - `{name} off`, `{name} away`, etc. for full vacation.
-# - `{name} available 4 hours/day`, `{name} 3h`, etc. for reduced availability.
-# FIXME: It doesn't currently support people with the same name, so, if needed, we'll need to change it.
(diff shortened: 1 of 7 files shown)


farhaanbukhsh (approved):
👍 

- [x] I tested this: set up the environment and created events for Daniel, the reflection of days off etc. are proper
- [x] I read through the code
~~- [ ] I checked for accessibility issues~~
- [x] Includes documentation, there is a follow up PR on handbook
~~- [ ] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.~~

PR #68, opened 2021-01-26
Fix second case when sprints were not created in a chronological order

Agrendalath:
Jira orders sprints by their IDs, so if they were not created in chronological order, then we need to search for an active one.


swalladge (approved):
@Agrendalath :+1:

- [ ] I tested this: will be tested when deployed
- [x] I read through the code
- [ ] NA I checked for accessibility issues
- [x] Includes documentation
- [x] I made sure any change in configuration variables is reflected in the corresponding
    client's `configuration-secure` repository.

PR #69, opened 2021-02-28
[BB-3174] Automate estimation session

Agrendalath:
This adds estimation session automation for the [Agile Poker](https://marketplace.atlassian.com/apps/700473/agile-poker-for-jira-planning-estimation) Jira app.

# Prerequisites
1. Set `.env` (you can find them [here](https://vault.opencraft.com:8200/ui/vault/secrets/secret/show/core/Sprints%20Development%20Envs)). There were some recent changes.
1. To `sprints/dashboard/tasks.py:update_estimation_session_task`, add the following below `for cell in get_cells(conn):`:
	```python
	if cell.board_id != 26:
	    continue
	```
1. To `sprints/dashboard/tasks.py:update_estimation_session_task`, add the following below `all_member_keys = list(current_member_keys | member_keys)`:
	```python
	print(f"Members of {cell.name}: {all_member_keys}")
	print(f"Unestimated tickets from {cell.name}: {[conn.issue(issue.id).key for issue in issues if issue.key.startswith(cell.key)]}")
	```
1. To `sprints/dashboard/tasks.py:move_estimates_to_tickets_task`, add the following below `for cell in get_cells(conn):`:
	```python
	if cell.board_id != 26:
	    continue
	```
1. To `sprints/dashboard/tasks.py:move_estimates_to_tickets_task`, add the following below `final_vote = get_poker_session_final_vote(votes, vote_values)`:
	```python
	print(f"The final vote for {conn.issue(issue).key} is {final_vote}.")
	```
1. To `sprints/dashboard/automation.py:ping_users_on_ticket`, add the following above `if not settings.DEBUG:`:
	```python
	print(message)
	```

# Testing instructions
1. Start the backend with `docker-compose -f local.yml up --build`.
1. Run migrations with `docker-compose -f local.yml exec django python manage.py migrate`.
1. Open Django shell with `docker-compose -f local.yml exec django python manage.py shell_plus` and run:
	```python
	from sprints.dashboard.tasks import *
	from sprints.dashboard.libs.jira import *
	
    conn = CustomJira(
        server=settings.JIRA_SERVER,
        basic_auth=(settings.JIRA_USERNAME, settings.JIRA_PASSWORD),
        options={
            'agile_rest_path': GreenHopperResource.AGILE_BASE_REST_PATH,
            'headers': {
                'Referer': 'https://tasks.opencraft.com/download/resources/com.spartez.jira.plugins.jiraplanningpoker/frontend/index.html',
            }
        }
    )
	```

[...]

PR #70, opened 2021-03-05
[BB-3174] Fix headers in API calls

Agrendalath:
After deploying #69, which added the custom `Referer` header, it turned out that the `content-type` is no longer set correctly, resulting in HTTP 415 response from some endpoints. This adds the missing header, which is enough for our use cases - we are not using Jira endpoints with content types different than `application/json`.


0x29a (approved):
:+1:

- [x] I read through the code
- [ ] I didn't test this since the change is trivial and @Agrendalath already tested it locally.

PR #73, opened 2021-03-12
[Hotfix] Fix assigning users to poker sessions, add task filtering

Agrendalath:
This fixes a case of having a different list of participants and scrum masters - e.g. for the bot account, which should not be a participant of an estimation session.

It also adds task filtering for the estimation session.


0x29a (approved):
:+1:

- [ ] I haven't tested this since the changes are trivial.
- [x] I carefully read through the code.
- [x] Includes documentation.

PR #74, opened 2021-03-12
[Hotfix] Fix updating Jira poker sessions

Agrendalath:
This fixes the HTTP 500 response from Jira while updating the session for some cells. It's related to undefined handling of user names vs user keys by Jira.


Agrendalath:
I tested this locally. Merging as a trivial change, because it only adds `?withUserKeys=true` to poker session requests.

PR #75, opened 2021-03-15
[Hotfix] Do not send emails on the estimation session update

Agrendalath:
It has been sending an email in every task run.


Agrendalath:
Merging as a trivial change.

PR #77, opened 2021-03-16
[BB-3174] Disable Jira session notifications v3

Agrendalath:
This is the third approach, as it does not use the native Jira API way of handling this.


Agrendalath:
Merging as a small change.

PR #78, opened 2021-03-22
[BB-2276] Fix retrieving Jira Poker configurations

Agrendalath:
This adds handling non-numeric values, which are inconsistently included in some configurations.

### DIFF
--- a/sprints/dashboard/libs/jira.py
+++ b/sprints/dashboard/libs/jira.py
@@ -314,11 +314,7 @@ class CustomJira(JIRA):
         :return: A list of possible vote values.
         """
         response = self._get_json(f'board/{board_id}/settings', base=self.AGILE_POKER_URL)
-        return [
-            float(vote['value'])
-            for vote in response.get('voteValues', [])
-            if vote['value'].replace('.', '', 1).isdigit()
-        ]
+        return [float(vote['value']) for vote in response.get('voteValues', []) if is_number(vote['value'])]
 
     def poker_session_results(self, session_id: int) -> dict[str, dict[str, dict[str, object]]]:
         """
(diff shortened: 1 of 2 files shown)


sprints/dashboard/libs/jira.py
```diff
@@ -314,8 +314,11 @@ def poker_session_vote_values(self, board_id: int) -> list[float]:
         :return: A list of possible vote values.
         """
         response = self._get_json(f'board/{board_id}/settings', base=self.AGILE_POKER_URL)
-        print(response)
-        return [float(vote['value']) for vote in response.get('voteValues', [])]
+        return [
+            float(vote['value'])
+            for vote in response.get('voteValues', [])
+            if vote['value'].replace('.', '', 1).isdigit()
```
swalladge, line 320: This feel like a hacky heuristic. It will fail in cases like `1..1`.  Perhaps something like this will be more reliable but still allow a nice list comprehension:

```
def is_number(n: str) -> bool:
  try:
    float(n)
  except:
    return False
  return True

... [float(n) for n in values if is_number(n)]
```

sprints/dashboard/tasks.py
```diff
@@ -595,11 +595,8 @@ def move_estimates_to_tickets_task() -> None:
                     votes = []
                     for result in results.values():
                         vote = result.get("selectedVote")
-                        try:
+                        if str(vote).replace('.', '', 1).isdigit():  # Ignore non-numeric answers.
```
swalladge, line 598: Same here; the try/except is probably more reliable

swalladge (changes requested):

swalladge:
@Agrendalath don't forget to add a ticket id to the PR title :)

sprints/dashboard/libs/jira.py
swalladge: Oh wait, my bad, missed that this was replacing it only once.  :grimacing: 

Still, this kind of indicates that it's not clear exactly what it's doing and how.

swalladge (approved):
@Agrendalath :+1:

- [x] I tested this: comms on mattermost; this has already been tested according to @Agrendalath 
- [x] I read through the code
- [x] I checked for accessibility issues
- [x] Includes documentation
- [ ] NA I made sure any change in configuration variables is reflected in the corresponding
    client's `configuration-secure` repository.

PR #81, opened 2021-05-17
[BB-4548] Unify sprint start/end dates

Agrendalath:
This unifies sprint start/end dates by using the higher-level API instead of the native Jira one. The also allows separating the Sprints logic from the Jira-specific features. It's the first step towards creating an abstraction, which is going to be used by [plugins](https://gitlab.com/opencraft/dev/sprints/-/blob/master/docs/discoveries/2.%20Sprints%20v2%20UX.md#12-plugins).

It also adds some tests.

Fixes https://gitlab.com/opencraft/dev/sprints/-/issues/16 (the content of this issue is a little different, so I've added a comment to explain the approach in `complete_sprint_task`).

This is going to be tricky to test manually, as it would require sprint manager permissions + manual API calls, as our Jira version has a very limited way of editing sprint details. Therefore, we can rely on the tests here.

## Testing instructions
1. Check if new tests make sense.
2. Check if the CI passes.

### DIFF
--- a/sprints/dashboard/tasks.py
+++ b/sprints/dashboard/tasks.py
@@ -281,13 +281,11 @@ def complete_sprint_task(board_id: int) -> None:
             conn.add_issues_to_sprint(next_sprint.id, issue_keys)
 
             # Open the next sprint.
-            start_date = datetime.now()
-            end_date = datetime.now() + timedelta(days=settings.SPRINT_DURATION_DAYS)
             conn.update_sprint(
                 next_sprint.id,
                 name=next_sprint.name,
-                startDate=start_date.isoformat(),
-                endDate=end_date.isoformat(),
+                startDate=next_sprint.startDate,
+                endDate=next_sprint.endDate,
                 state='active',
             )
 
--- a/sprints/dashboard/utils.py
+++ b/sprints/dashboard/utils.py
@@ -503,10 +503,9 @@ def create_next_sprint(conn: CustomJira, sprints: List[Sprint], cell_key: str, b
     end_date_str = _get_sprint_end_date(start_date_str)
 
     sprint_number = get_sprint_number(last_sprint) + 1
-    name_date = start_date.strftime(settings.JIRA_API_DATE_FORMAT)
 
     return conn.create_sprint(
-        name=f'{cell_key}.{sprint_number} ({name_date})',
+        name=f'{cell_key}.{sprint_number} ({start_date_str})',
         board_id=board_id,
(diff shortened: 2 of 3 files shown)


sprints/dashboard/tasks.py
```diff
@@ -281,13 +281,11 @@ def complete_sprint_task(board_id: int) -> None:
...
             # Open the next sprint.
-            start_date = datetime.now()
-            end_date = datetime.now() + timedelta(days=settings.SPRINT_DURATION_DAYS)
             conn.update_sprint(
                 next_sprint.id,
                 name=next_sprint.name,
-                startDate=start_date.isoformat(),
-                endDate=end_date.isoformat(),
+                startDate=next_sprint.startDate,
+                endDate=next_sprint.endDate,
```
Agrendalath, line 288: These are already set by the `create_next_sprint` function, so they don't need to be regenerated.

sprints/dashboard/utils.py
```diff
@@ -497,23 +497,26 @@ def create_next_sprint(conn: CustomJira, sprints: List[Sprint], cell_key: str, b
...
 
-    future_next_sprint_number = get_sprint_number(last_sprint) + 1
-    future_name_date = (end_date + timedelta(days=1)).strftime(settings.JIRA_API_DATE_FORMAT)
-    future_end_date = end_date + timedelta(days=settings.SPRINT_DURATION_DAYS)
+    start_date = parse(get_sprint_end_date(last_sprint)) + timedelta(days=1)
+    start_date_str = start_date.strftime(settings.JIRA_API_DATE_FORMAT)
+    end_date_str = _get_sprint_end_date(start_date_str)
+
+    sprint_number = get_sprint_number(last_sprint) + 1
+    name_date = start_date.strftime(settings.JIRA_API_DATE_FORMAT)
```
arjunsinghy96, line 506: This contains the same value as `start_date_str`. Do we really need it?

arjunsinghy96 (approved):
:+1: @Agrendalath LGTM. I have left the nit though. Could you please address that before merging this?

- [x] I tested that the CI has passed and the new tests LGTM
- [x] I read through the code
- [ ] ~~I checked for accessibility issues~~
- [x] Includes documentation
- [ ] ~~I made sure any change in configuration variables is reflected in the corresponding~~ client's `configuration-secure` repository.

sprints/dashboard/utils.py
Agrendalath: Not at all. Good catch.

PR #83, opened 2021-08-03
BB-3992: Change logic of estimated reviewer time on tasks

alfredchavez:
This changes the logic of review time estimation for tasks, according to the story points assignated to a task.
Here the estimations are defined on an environment variable

Ticket: https://gitlab.com/opencraft/dev/sprints/-/issues/2

Test Instructions:
- Check that unit tests makes sense(and cover correctly the implemented logic)
- I created some dummy tickets to test the review time estimation
-- Create and set an `.env` file (like [this](https://vault.opencraft.com:8200/ui/vault/secrets/secret/show/core/Sprints%20Development%20Envs)) and add the following environment variable
`SPRINT_HOURS_RESERVED_FOR_REVIEW={"null": 2, "1.9": 0.5, "2": 1, "3": 2, "5": 3, "5.1": 5}`
--Start the backend with `docker-compose -f local.yml up --build`
-- Run migrations with `docker-compose -f local.yml exec django python manage.py migrate`.(if you are running it for the first time)
-- Open the django shell with  `docker-compose -f local.yml exec django python manage.py shell_plus` to execute the following code snippets
-- Run this snippet so to create a Mocked/Fake Dashboard, to use its methods on our test
``` python
conn = CustomJira(
    server=settings.JIRA_SERVER,
    basic_auth=(settings.JIRA_USERNAME, settings.JIRA_PASSWORD),
    options={
        'agile_rest_path': GreenHopperResource.AGILE_BASE_REST_PATH,
        'headers': {
            'Referer': 'https://tasks.opencraft.com/download/resources/com.spartez.jira.plugins.jiraplanningpoker/frontend/index.html',
        }
    }
)

dashboard = Dashboard(26, conn)
```
-- Then run this snippet to prepare the required parameters for DashboardIssue class
``` python
quickfilters: List[QuickFilter] = conn.quickfilters(26)
members = get_cell_members(quickfilters)
sprints = get_all_sprints(conn, 26)
active_sprint_ids = {sprint.id for sprint in sprints['active']}
```
-- Run this to get the `dummy issues` we are going to use, along with the expected "review time" of each one
``` python
issues = [
    (conn.issue("BB-4660"), 2.0 * 3600),
    (conn.issue("BB-4661"), 0.5 * 3600),
    (conn.issue("BB-4662"), 1.0 * 3600),
    (conn.issue("BB-4663"), 2.0 * 3600),
    (conn.issue("BB-4664"), 3.0 * 3600),
    (conn.issue("BB-4665"), 5.0 * 3600),
    (conn.issue("BB-4666"), 3.0 * 3600)]
```
-- Finally run this to verify the method is behaving as expected for each dummy issue
``` python
for issue, expected_time_seconds in issues:
    dissue = DashboardIssue(issue, active_sprint_ids, members, dashboard.unassigned_user, dashboard.other_cell, issue_fields, dashboard.cell.key)
    assert dissue.review_time == expected_time_seconds
```


config/settings/base.py
```diff
@@ -609,6 +609,18 @@
     },
 }
 
+# Time estimates for reviewing tasks based on the assigned story points
+#
+# SPRINT_HOURS_RESERVED_FOR_REVIEW {
+#     "0": Review time if task is not estimated,
```
Agrendalath, line 615: This is not true. Please make [this GET request](https://tasks.opencraft.com/rest/api/2/issue/BB-4594) and check the value of `customfield_10002` ("Story Points", as described [here](https://tasks.opencraft.com/rest/api/2/field)). It is `null`.
We could handle this by adding `null` as a dict key.

README.rst
```diff
@@ -29,7 +29,13 @@ The basic idea for calculating estimations is the following:
...
 1. `SPRINT_HOURS_RESERVED_FOR_MEETINGS` hours are reserved for the meetings for each sprint.
 2. `SPRINT_HOURS_RESERVED_FOR_EPIC_MANAGEMENT` hours are reserved for epic management for each sprint.
-3. 1 hour is planned for reviewing each task with <= 3 story points. For bigger tasks, 2 hours are reserved.
+3. Reserved time for reviewing is defined in `SPRINT_HOURS_RESERVED_FOR_REVIEW`, where you can define the review time for:
+    * Less than 2 story points("<2"), default is 0.5 hours
+    * 2 story point("2"), default is 1 hour
+    * 3 story points("3"), default is 2 hours
+    * 5 story points("5"), default is 3 hours
+    * More than 5 story points(">5"), defaults to 5 hours
+    * When a task is not estimated("0"), defaults to 2 hours
```
Agrendalath, line 38: I see that it's an empty `dict` by default in the settings. Am I missing something?

sprints/dashboard/models.py
```diff
@@ -124,12 +124,36 @@ def assignee_time(self) -> int:
...
+        if self.story_points == 2:
+            return story_points_hours_map.get("2")
+
+        if self.story_points == 3:
+            return story_points_hours_map.get("3")
+
+        if self.story_points == 5:
+            return story_points_hours_map.get("5")
+
+        return story_points_hours_map.get(">5")
```
Agrendalath, line 148: It's not a flexible approach. Instead of hardcoding these values, we could define that undefined story points will be applied to the nearest defined ones. E.g.:
```python
{
    None: 2
    1.9: 0.5,
    2: 1,
    3: 2,
    5: 3,
    5.1: 5,
}
```
In this case, `0SP == 0.5h, 1SP == 0.5h, 1.99SP == 1h, 999SP == 5h, None == 2h`.

What do you think?

sprints/dashboard/models.py
```diff
@@ -140,9 +164,10 @@ def review_time(self) -> int:
             if self.is_epic or self.status in settings.SPRINT_STATUS_NO_MORE_REVIEW:
                 return 0
 
-            if self.story_points <= 3:
-                return SECONDS_IN_HOUR
-            return int(2 * SECONDS_IN_HOUR)
+            review_time = self.get_review_time_from_story_points()
+            if review_time is None:  # if some value is not defined on the `SPRINT_HOURS_RESERVED_FOR_REVIEW` setting
+                raise ValueError
```
Agrendalath, line 169: Is this exception handled? If not, it's going to break the dashboard. See my previous comment - we probably don't need this.

Agrendalath (changes requested):
@alfredchavez, I've left some comments.

Also, please add the proper testing instructions. This is very easy to test by creating dummy tickets. Example instructions: #65, #69, #59.

sprints/dashboard/models.py
alfredchavez: Yep, I had my questions on this, now i realize that I misunderstood the example approach :) haha. ok, I'll work on that

config/settings/base.py
alfredchavez: Yes, now I read a null key from the dict

README.rst
alfredchavez: Changed the default to meet the defined allocation logic in the issue description

sprints/dashboard/models.py
alfredchavez: Changed the logic, so now goes to the closest/nearest defined  one

sprints/dashboard/models.py
alfredchavez: Removed the exception

alfredchavez:
@Agrendalath I still have a doubt, the situation in which an empty dict(no null and num value) is pretty much not handled, I added to the logic that if no num value is defined it will go to the "null" number of hours, should I enforce on base.py maybe that at least "null" should be defined(added to the README also)?

README.rst
```diff
@@ -29,7 +29,23 @@ The basic idea for calculating estimations is the following:
...
+   .. code::
+
+        SPRINT_HOURS_RESERVED_FOR_REVIEW {
+            "null": 2,
+            "1.9": 1,
+            "2": 3,
+            "5.1": 6
+        }
+
+   Here we are defining that if no story point is set for an issue, it will reserve 2 hours, for 1.9 story points it will reserve 1 hour,
```
Agrendalath, line 43:
```suggestion
   Here we are defining that:

   1.  If no story points are set for an issue, it will reserve 2 hours.
   2.  For 1.9 story points it will reserve 1 hour.
```

Etc…
It will be more readable.

config/settings/base.py
```diff
@@ -609,6 +609,20 @@
     },
 }
 
+# Time estimates for reviewing tasks based on the assigned story points
+#
+# SPRINT_HOURS_RESERVED_FOR_REVIEW {
```
Agrendalath, line 614:
```suggestion
{
```

The variable name looks redundant here.

config/settings/base.py
```diff
@@ -609,6 +609,20 @@
     },
 }
 
+# Time estimates for reviewing tasks based on the assigned story points
```
Agrendalath, line 612:
```suggestion
# Time estimates for reviewing tasks based on the assigned story points. Configuration format:
```

config/settings/base.py
```diff
@@ -609,6 +609,20 @@
...
+#     "1.9": Review time if task has less than 2 story points,
+#     "2": Review time if task has 2 story points,
+#     "3": Review time if task has 3 story points,
+#     "5": Review time if task has 5 story points,
+#     "5.1": Review time if task has more than 5 story points
+# }
+# Any time estimate that is not defined here, will use the "review time" from the closest value defined
+
+SPRINT_HOURS_RESERVED_FOR_REVIEW = json.loads(env.str("SPRINT_HOURS_RESERVED_FOR_REVIEW",
+                                                      '{"null": 2, "1.9": 0.5, "2": 1, "3": 2, "5": 3, "5.1": 5}'))
```
Agrendalath, line 625: Nit.

```suggestion
SPRINT_HOURS_RESERVED_FOR_REVIEW = json.loads(
    env.str("SPRINT_HOURS_RESERVED_FOR_REVIEW", '{"null": 2, "1.9": 0.5, "2": 1, "3": 2, "5": 3, "5.1": 5}')
)
```

sprints/dashboard/models.py
```diff
@@ -84,9 +84,9 @@ def __init__(
             # Possible for epics
             self.current_sprint = False
         try:
-            self.story_points = int(getattr(issue.fields, issue_fields['Story Points'], 0))
+            self.story_points = getattr(issue.fields, issue_fields['Story Points'], None)
         except (AttributeError, TypeError):
```
Agrendalath, line 88: I suppose we don't need this exception anymore, as we're not casting the value?

config/settings/base.py
Agrendalath, line 625: Wouldn't it make sense to use actual values instead of strings as keys? I know that it's a limitation of JS, but there are two workarounds:
1. Create something like this:
    ```python
    def jsonKeys2int(x):
        if isinstance(x, dict):
            res = {}
            for k, v in x.items():
                try:
                    new_key = float(k)
                except ValueError:
                    new_key = None
                res[new_key] = v
            return res
        return x
     ```
    And pass it as `object_hook` to `json.loads`.
1. Use the following JSON structure for this configuration:
    ```json
    [
        [null, 2],
        [1.9, 0.5],
        [2,1],
        [3,2],
        [5,3],
        [5.1, 5]
    ]
    ```
    And then simply convert it with `dict()`.

I believe that the first approach is cleaner, though. What do you think?

.envs/.local/.django
```diff
@@ -20,6 +20,9 @@ JIRA_PASSWORD=
 JIRA_CELL_ROLE_ACCOUNT=
 JIRA_CELL_ROLES=
 
+#Sprint
+SPRINT_HOURS_RESERVED_FOR_REVIEW=
```
Agrendalath, line 24: This is going to break the environment when `SPRINT_HOURS_RESERVED_FOR_REVIEW` is not defined in envs.
```suggestion
# SPRINT_HOURS_RESERVED_FOR_REVIEW=
```

.envs/.production/.django
```diff
@@ -54,6 +54,9 @@ JIRA_PASSWORD=
 JIRA_CELL_ROLE_ACCOUNT=
 JIRA_CELL_ROLES=
 
+# Sprint
+SPRINT_HOURS_RESERVED_FOR_REVIEW=
```
Agrendalath, line 58: This is going to break the environment when `SPRINT_HOURS_RESERVED_FOR_REVIEW` is not defined in envs.
```suggestion
# SPRINT_HOURS_RESERVED_FOR_REVIEW=
```

.envs/.local/.django
```diff
@@ -20,6 +20,9 @@ JIRA_PASSWORD=
 JIRA_CELL_ROLE_ACCOUNT=
 JIRA_CELL_ROLES=
 
+#Sprint
```
Agrendalath, line 23: Nit: missing space.
```suggestion
# Sprint
```

sprints/dashboard/models.py
```diff
@@ -124,12 +124,36 @@ def assignee_time(self) -> int:
 
         return max(self.time_estimate - self.review_time, 0)  # We don't want negative values here.
 
+    def get_review_time_from_story_points(self):
+        issue_story_points = self.story_points
+        review_hours_by_story_points = settings.SPRINT_HOURS_RESERVED_FOR_REVIEW
+        if issue_story_points is None:
+            return review_hours_by_story_points.get("null")
```
Agrendalath, line 131: Or something along these lines.
```suggestion
        if (hours := review_hours_by_story_points.get(issue_story_points)) is not None:
            return hours
```

sprints/dashboard/models.py
```diff
@@ -124,12 +124,36 @@ def assignee_time(self) -> int:
...
+                continue
+            story_point_num_value = float(story_points)
+            value_difference = abs(story_point_num_value - issue_story_points)
+            if nearest_value == "null" or \
+               (value_difference == minimum_value_difference and story_point_num_value > float(nearest_value)) or \
+               value_difference <= minimum_value_difference:
+                minimum_value_difference = value_difference
+                nearest_value = story_points
+
+        return review_hours_by_story_points.get(nearest_value)
```
Agrendalath, line 148: This looks quite complex. Why don't we just do something like this?
```python
# Some explanation that the value is not defined, so we're getting the closest one - this should also go to the docstring of this function.
defined_story_points = sorted([key for key in review_hours_by_story_points.keys() if key is not None])
nearest_story_points min(defined_story_points, key=lambda x:abs(x - self.story_points))
return review_hours_by_story_points[nearest_story_points]
```

Agrendalath (changes requested):
@alfredchavez, I did the second pass and left some comments. Also, please don't squash the commits during the review, as this makes tracking changes harder. Additionally, it invalidates code review comments.

> I still have a doubt, the situation in which an empty dict(no null and num value) is pretty much not handled, I added to the logic that if no num value is defined it will go to the "null" number of hours, should I enforce on base.py maybe that at least "null" should be defined(added to the README also)?

Yes, we should definitely verify that this value is defined and add it to the README.

README.rst
```diff
@@ -29,7 +29,23 @@ The basic idea for calculating estimations is the following:
...
 1. `SPRINT_HOURS_RESERVED_FOR_MEETINGS` hours are reserved for the meetings for each sprint.
 2. `SPRINT_HOURS_RESERVED_FOR_EPIC_MANAGEMENT` hours are reserved for epic management for each sprint.
-3. 1 hour is planned for reviewing each task with <= 3 story points. For bigger tasks, 2 hours are reserved.
+3. Reserved time for reviewing is defined in `SPRINT_HOURS_RESERVED_FOR_REVIEW` as a json dict, where you can define the review time for any
+   amount of story points, and also if no story points are defined("null") for a Jira issue. e.g.
+   .. code::
+
+        SPRINT_HOURS_RESERVED_FOR_REVIEW {
+            "null": 2,
+            "1.9": 1,
```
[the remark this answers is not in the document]
alfredchavez: Added it to the README example

sprints/dashboard/models.py
alfredchavez: Yep

config/settings/base.py
alfredchavez: Yes, I think the first approach looks better

sprints/dashboard/models.py
alfredchavez: Sure, that looks way simple

config/settings/base.py
```diff
@@ -609,6 +609,42 @@
...
+            try:
+                processed_key = float(key)
+            except ValueError:
+                processed_key = None
+            processed_json_dict[processed_key] = value
+        return processed_json_dict
+    return json_dict
+
+
+def preprocess_sprints_hours_reserved_for_review(json_dict):
```
alfredchavez, line 626: btw, created this method were I "enforce" the null value being defined and also casts the strings keys to float values(and None)

README.rst
```diff
@@ -30,21 +30,34 @@ The basic idea for calculating estimations is the following:
...
+   4. If 2 story points are set for an issue, it will reserve 3 hours for its review.
+   5. If 5.1 story points are set for an issue, it will reserve 6 hours for its review.
+   6. If the ticket has an amount of story points that is not defined in the `SPRINT_HOURS_RESERVED_FOR_REVIEW` setting,
+      then it will use the review time for the nearest number of story point defined. For example:
+       - If 3 story points are set for an issue, it will reserve, 3 hours for its review, as the closest number of story points
+         defined is 2
+       - If more than 5.1 story points (6, 10, 20, etc.) are defined for an issue, it will reserve 6 hours for its review,
+         because the closest story points value defined is 5.1
+
+   **The "null" value is REQUIRED to be defined in `SPRINT_HOURS_RESERVED_FOR_REVIEW`.**
```
Agrendalath, line 60:
```suggestion
   **Note**: The ``null`` value is **required** in ``SPRINT_HOURS_RESERVED_FOR_REVIEW``.
```

config/settings/base.py
```diff
@@ -609,9 +610,29 @@
...
+    if "null" not in json_dict:
+        raise ValueError('required "null" key is missing for SPRINT_HOURS_RESERVED_FOR_REVIEW')
```
Agrendalath, line 630: This should raise [ImproperlyConfigured](https://docs.djangoproject.com/en/3.2/ref/exceptions/#improperlyconfigured).
```suggestion
        raise ImproperlyConfigured('Required "null" key is missing from "SPRINT_HOURS_RESERVED_FOR_REVIEW".')
```

config/settings/base.py
Agrendalath:
```suggestion
def validate_sprints_hours_reserved_for_review(json_dict):
```
What do you think?

config/settings/base.py
```diff
@@ -609,9 +610,29 @@
...
+            try:
+                processed_key = float(key)
+            except ValueError:
+                processed_key = None
+            processed_json_dict[processed_key] = value
+        return processed_json_dict
+    return json_dict
+
+
+def preprocess_sprints_hours_reserved_for_review(json_dict):
```
Agrendalath, line 628: Please add the docstring.

config/settings/base.py
```diff
@@ -609,9 +610,29 @@
     },
 }
 
-# Time estimates for reviewing tasks based on the assigned story points
-#
-# SPRINT_HOURS_RESERVED_FOR_REVIEW {
+
+def json_keys_to_float(json_dict):
```
Agrendalath, line 614: Please add the docstring.

config/settings/base.py
```diff
@@ -621,8 +642,11 @@
 # }
 # Any time estimate that is not defined here, will use the "review time" from the closest value defined
 
-SPRINT_HOURS_RESERVED_FOR_REVIEW = json.loads(env.str("SPRINT_HOURS_RESERVED_FOR_REVIEW",
-                                                      '{"null": 2, "1.9": 0.5, "2": 1, "3": 2, "5": 3, "5.1": 5}'))
+
+SPRINT_HOURS_RESERVED_FOR_REVIEW = preprocess_sprints_hours_reserved_for_review(json.loads(
+    env.str("SPRINT_HOURS_RESERVED_FOR_REVIEW", '{"null": 2, "1.9": 0.5, "2": 1, "3": 2, "5": 3, "5.1": 5}')
+))
```
Agrendalath, line 648: Nit: wouldn't using [object_hook](https://docs.python.org/3/library/json.html) be clearer than adding a new level of the function calls?

sprints/dashboard/models.py
```diff
@@ -85,7 +85,7 @@ def __init__(
             self.current_sprint = False
         try:
             self.story_points = getattr(issue.fields, issue_fields['Story Points'], None)
-        except (AttributeError, TypeError):
+        except AttributeError:
```
Agrendalath, line 88: We don't need `AttributeError` as well - we're defining the default value here, no?

sprints/dashboard/models.py
```diff
@@ -124,34 +124,34 @@ def assignee_time(self) -> int:
 
         return max(self.time_estimate - self.review_time, 0)  # We don't want negative values here.
 
-    def get_review_time_from_story_points(self):
+    def calculate_review_time_from_story_points(self) -> float:
+        """
+        Calculate time needed for a review, by using the story points defined in the issue.
+
+        If the story points defined in the issue, is defined in SPRINT_HOURS_RESERVED_FOR_REVIEW, then
```
Agrendalath, line 131:
```suggestion
        If the story points set in the issue are defined in `SPRINT_HOURS_RESERVED_FOR_REVIEW`, then
```

sprints/dashboard/models.py
```diff
@@ -124,34 +124,34 @@ def assignee_time(self) -> int:
...
+        it is going to return the review time for that value,
```
Agrendalath, line 132:
```suggestion
        it is going to return the review time for that value.
```

sprints/dashboard/models.py
```diff
@@ -124,34 +124,34 @@ def assignee_time(self) -> int:
...
+        otherwise will use the review time of the closest value defined in SPRINT_HOURS_RESERVED_FOR_REVIEW.
```
Agrendalath, line 133:
```suggestion
        Otherwise, it will use the review time of the closest value defined in `SPRINT_HOURS_RESERVED_FOR_REVIEW`.
```

Agrendalath (changes requested):

sprints/dashboard/models.py
alfredchavez: yep

config/settings/base.py
alfredchavez: yes, didn't know about it, it will make it more clean

Agrendalath (approved):
:+1:

* [x]  I tested this: tested this locally
* [x]  Includes documentation
* [x] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository: n/a

